PR #15, opened 2018-04-05
feat(type): added path

martintamare:
Path give the hability to do things such as

```
@api.handles(
      rule='/vault/<int:service_id>/list/<path:path>',
      method='GET',
)
```

Which validate : `/vault/2/list/secret`
but also `/vault/2/list/secret/toto/titit/tata`

It comes from http://flask.pocoo.org/snippets/57/


barakalon:
@martintamare This change should now be available in version 1.0.5

martintamare:
Awesome, thanks for the reactivity !

PR #16, opened 2018-07-24
Pass flask Response objects straight through if returned by view functions

mjbryant:
The test shows how this can be used to, say, override default Flask/werkzeug/rebar processing. Sometimes you just to control what gets returned, you know?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,9 @@ man/
 
 # pytest
 .pytest_cache
+
+# pyenv
+.python-version
+
+# ctags
+tags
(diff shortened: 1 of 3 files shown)


mjbryant:
As @tmlippman pointed out, I could have done this before, since `rv` is passed through if no marshal schema is defined. This wouldn't work if I wanted to define a marshal schema for one status code and not another, but that's a finer-grained use case that I don't care about for the moment.

barakalon:
Yeah, I think the response handling is one of flask-rebar's biggest weaknesses at the moment - it doesn't adapt to atypical use cases very well.

I'd be super interested to hear ideas for improvement. An elegant solution for custom responses/content negotiation that plays well with swagger would be sweet.

Issue #24, opened 2018-10-26
_init_error_handling() breaks Flask's handling of RequestRedirect when not in debug mode


twosigmajab:
Out of the box, Flask sets up nice automatic handling of trailing slash redirects. So if you have a handler mapped to a URL like `/swagger/ui/` (such as the one Flask-Rebar sets up for you), but the user makes a request like `/swagger/ui` (note no trailing slash), then normally, Flask serves the user a redirect to `/swagger/ui/` without your having to do anything extra, *whether or not `app.debug` is True*.

However, Flask-Rebar's `_init_error_handling` is interfering with this in the case that `app.debug` is False. I hit this after writing a function `start_devserver(debug=False)`. Before calling `app.run(debug=debug)`, it prints out the string `f"* Swagger UI available at http://localhost:5000/{registry.prefix}/{registry.swagger_ui_path}"`. With `debug=False`, clicking the resulting link would result in a 500 error and the following stack trace:
```
[2018-10-26 15:57:55,690] ERROR in app: Exception on /swagger/ui [GET]
Traceback (most recent call last):
  File "/private/tmp/testvenv/lib/python3.7/site-packages/werkzeug/routing.py", line 1538, in match
    rv = rule.match(path, method)
  File "/private/tmp/testvenv/lib/python3.7/site-packages/werkzeug/routing.py", line 776, in match
    raise RequestSlash()
werkzeug.routing.RequestSlash

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/private/tmp/testvenv/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/private/tmp/testvenv/lib/python3.7/site-packages/flask/app.py", line 1791, in dispatch_request
    self.raise_routing_exception(req)
  File "/private/tmp/testvenv/lib/python3.7/site-packages/flask/app.py", line 1774, in raise_routing_exception
    raise request.routing_exception
  File "/private/tmp/testvenv/lib/python3.7/site-packages/flask/ctx.py", line 336, in match_request
    self.url_adapter.match(return_rule=True)
  File "/private/tmp/testvenv/lib/python3.7/site-packages/werkzeug/routing.py", line 1542, in match
    safe='/:|+') + '/', query_args))
werkzeug.routing.RequestRedirect: 301 Moved Permanently: None
[...]

PR #25, opened 2018-10-29
Handle RequestRedirect errors properly

barakalon:
A proposed fix for https://github.com/plangrid/flask-rebar/issues/24


barakalon:
Looks like this is incompatible with Flask 0.10.*... working on it...

flask_rebar/rebar.py
```diff
@@ -47,6 +50,14 @@
 ])
 
 
+# To catch redirection exceptions, app.errorhandler expects 301 in versions
+# below 0.11.0, but the exception itself in version greater than 0.11.0.
+if version.parse(flask_version) < version.parse('0.11.0'):
+    redirect_error = 301
```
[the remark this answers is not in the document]
barakalon: Ah, sure.

tests/test_rebar.py
```diff
@@ -508,3 +509,21 @@ def force_500():
...
+
+        app = create_rebar_app(rebar)
+
+        app.debug = False
+        resp = app.test_client().get(path='/with_trailing_slash')
+        self.assertEqual(resp.status_code, 301)
+        self.assertEqual(resp.content_type, 'application/json')
+        self.assertTrue(resp.headers['Location'].endswith('/with_trailing_slash/'))
+
+        app.debug = True
```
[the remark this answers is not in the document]
barakalon: I think flask-rebar should reliably return JSON when not in debug mode, which is why the behavior is a little different than flask's built in behavior.

It _does_ set the `Location` header properly, so browsers will still automatically redirect.

flask_rebar/rebar.py
```diff
@@ -15,11 +15,14 @@
 from collections import namedtuple
 from copy import copy
 from functools import wraps
+from packaging import version
```
brockhaywood: For some reason, i need to install `packaging` specifically using pip3 in python 3.5

barakalon: Ah, crap. I should've used another way of doing this comparison

barakalon: Let me change this real quick

barakalon: https://github.com/plangrid/flask-rebar/pull/29

PR #26, opened 2018-10-29
Improve marshal_schema / response header handling

twosigmajab:
- No longer ignore headers in tuple responses
- Allow passing headers to request_utils.response
- No longer jsonify responses that have a None schema

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -17,9 +17,7 @@ from copy import copy
 from functools import wraps
 
 import marshmallow
-from flask import current_app, jsonify
-from flask import g
-from flask import request
+from flask import current_app, g, jsonify, make_response, request
 
 from flask_rebar import messages
 from flask_rebar.authenticators import USE_DEFAULT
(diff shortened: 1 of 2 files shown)


flask_rebar/rebar.py
```diff
@@ -83,36 +81,20 @@ def wrapped(*args, **kwargs):
...
-        if marshal_schema:
-            if isinstance(rv, tuple):
-                data, status_code = rv[0], rv[1]
-            else:
-                data, status_code = rv, 200
-
-            try:
-                schema = marshal_schema[status_code]
-            except KeyError:
-                raise
```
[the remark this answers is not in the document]
barakalon: You're correct - this is better. I bet there used to be behavior in the except block, but now its gone.

flask_rebar/rebar.py
```diff
@@ -83,36 +81,30 @@ def wrapped(*args, **kwargs):
 
         rv = f(*args, **kwargs)
 
-        if marshal_schema:
-            if isinstance(rv, tuple):
-                data, status_code = rv[0], rv[1]
```
twosigmajab: The existing behavior ignores any headers the user may have passed in the third element of the tuple, and also doesn't support passing a 2-tuple like `(body, headers)`, or a string status (see "tuple" under http://flask.pocoo.org/docs/1.0/api/#flask.Flask.make_response)

twosigmajab:
Meant to say, this PR was prompted by what I expect to be a fairly typical use case: wanting to return a 201 Created response with an empty body and a Location header containing the URL of the newly-created resource. Passing `marshal_schema: {201: None}` to Flask-Rebar successfully documents this intention, but Flask-Rebar currently swallows any headers you attempt to pass it in a tuple response.

twosigmajab:
Thanks for the quick review, @barakalon! I'm actually about to knock off for the day and am not sure when I'll have time to get back to adding the tests -- I'd love to pass it back to you at this point if that works for you. Thanks again!

flask_rebar/rebar.py
```diff
@@ -83,36 +81,30 @@ def wrapped(*args, **kwargs):
...
-                )
+        if not marshal_schema:
+            return rv
 
-            marshaled = marshal(
-                data=data,
-                schema=schema
-            )
+        data, status_code, headers = rv, 200, None
+        if isinstance(rv, tuple):
```
twosigmajab: Just found Flask's own code for handling tuple responses:
https://github.com/pallets/flask/blob/7e714bd28b6e96d82b2848b48cf8ff48b517b09b/flask/app.py#L1932

Would probably be better to match that more closely here, or maybe better yet just call it and then modify the result as needed.

barakalon:
No problems - I've picked up the baton here: https://github.com/plangrid/flask-rebar/pull/28

PR #32, opened 2018-10-30
Allow disabling OrderedDicts in generated swagger

twosigmajab:
If you try to `yaml.dump` a rebar-generated swagger object, you end up with yaml that looks like this:

```yaml
!!python/object/apply:collections.OrderedDict
- - - consumes
    - [application/json]
  - - definitions
    - !!python/object/apply:collections.OrderedDict
      - - - Error
          - !!python/object/apply:collections.OrderedDict
            - - - properties
                - !!python/object/apply:collections.OrderedDict
                  - - - errors
                      - !!python/object/apply:collections.OrderedDict
                        - - [type, object]
                    - - message
                      - !!python/object/apply:collections.OrderedDict
                        - - [type, string]
              - - required
                - [message]
              - [title, Error]
              - [type, object]
...
```

It looks like this is due to the current behavior of unconditionally converting dicts to OrderedDicts.

This PR allows passing `order_dicts=False` to `swagger_generator.generate(...)` to suppress the conversion. 

This may also be desired by users who want to avoid the extra work if they're using a Python implementation where dicts preserve insertion order already (e.g. PyPy or CPython >= 3.6).


flask_rebar/swagger_generation/swagger_generator.py
```diff
@@ -367,7 +367,8 @@ def generate(
             host=None,
             schemes=None,
             consumes=None,
-            produces=None
+            produces=None,
+            order_dicts=True,
```
barakalon: Nitpick, but I think this should be named `sort_keys`. This is subjective, but I feel like that's easier to understand as a user of this interface. Let me know if you disagree - my instincts are not always correct 😬 

It should also be added to the docstring.

twosigmajab: Done in e8dce7c. I actually had the same thought myself, but went with `order_dicts` to better match the existing `_recursively_order_dicts(...)` method. I changed that to `_recursively_sort_dicts` in e8dce7c as well. Make sense?

barakalon: Yep! that's better

barakalon:
@twosigmajab I think this branch needs to be updated before merging

juliusiv:
@twosigmajab it looks like the argument is still `order_dicts` but the docstring is `sort_keys`?

twosigmajab:
sorry about that, fixed now!

twosigmajab:
Just clicked "Update branch".

juliusiv:
thanks for the contribution @twosigmajab ! I'll try to get a release out today with an updated changelog.

twosigmajab:
Wonderful, thanks!

twosigmajab:
Hey @juliusiv, happy Monday! Hope my opening #33 didn't hold up doing the next release... Psyched for those changes :)

juliusiv:
@twosigmajab happy Monday to you as well! Version 1.1.0 should be released.

Issue #37, opened 2018-11-26
Prepare for Marshmallow 3


twosigmajab:
https://github.com/plangrid/flask-rebar/blob/master/setup.py#L19 specifies `marshmallow>=2.13`. Right now this is picking up Marshmallow 2, but as soon as Marshmallow 3 final comes out, this will start picking up v3, which has several backwards-incompatible changes:

https://marshmallow.readthedocs.io/en/3.0/changelog.html

Are any updates necessary before then?

barakalon:
Woof, I've running the tests with 3.0.0b20... and there are some pretty nasty breaking changes coming through.

We've got a couple of options:
1) modify the current dependency to be `marshmallow>=2.13,<3` and release a new major version of flask-rebar that depends on `marshmallow>=3`. This is probably the least convenient for consumers of this package, but will lead to much simpler code.
2) support both major version of marshmallow in flask-rebar. I'll keep hacking away at this... but it's not going down easy.

twosigmajab:
@barakalon Thanks for looking into this! Doing a new release with just the marshmallow dep pinned to `marshmallow>=2.13,<3` that users can upgrade to now seems like it should be done asap, no? That way users can upgrade now to make sure they don't one day start pulling in an incompatible marshmallow version and their apps suddenly start breaking.

twosigmajab:
(and in general, each dependency should be pinned within its semver-compatible range, in case there are others in the same boat)

barakalon:
I put up a PR that free the need to add `marshmallow>=2.13,<3` to requirements.txt.

It's a little messy, so I'm going to sleep on it and come back to it tomorrow.

> and in general, each dependency should be pinned within its semver-compatible range

Yep, agreed.

twosigmajab:
> I put up a PR that free the need to add `marshmallow>=2.13,<3` to requirements.txt.

Nice! Looks like it's #43 -- just left some comments there. Also, don't forget you'd still want to add the semver upper bound like `marshmallow>=2.13,<4` (as we just agreed) until compatibility with some future marshmallow v4 is proven.

And re requirements.txt see #44.

twosigmajab:
[...]

PR #43, opened 2018-11-27
Prepare for Marshmallow version 3

barakalon:
Addresses https://github.com/plangrid/flask-rebar/issues/37

Some ugly things here. Happy to receive feedback :)


flask_rebar/request_utils.py
```diff
@@ -10,6 +10,7 @@
 from __future__ import unicode_literals
 
 import copy
+from collections import Mapping
```
twosigmajab: On Python 3.7+ this will generate a deprecation warning. See https://github.com/jab/bidict/blob/master/bidict/compat.py#L41 
One more thing to move into the compat module.

flask_rebar/request_utils.py
```diff
@@ -18,6 +19,41 @@
...
+
+if MARSHMALLOW_MAJOR_VERSION == 2:
+    def _normalize_result(result):
+        return result.data
+else:
+    def _normalize_result(result):
+        return result
+
+
+class HeadersProxy(Mapping):
```
twosigmajab: Doesn't this already exist in Python 3? https://docs.python.org/3/library/types.html#types.MappingProxyType (and for supporting python 2 as well there's https://pypi.org/project/dictproxyhack/)

Another one-line approach (which is also natively supported in Python 2) could be to call `Mapping.register(werkzeug.datastructures.Headers)`, making use of https://docs.python.org/3/library/abc.html#abc.ABCMeta.register
(Maybe that's a bit too close to monkey-patching / leaking side effects into users' applications. Hard to imagine that ever biting someone, but that's what they all say.)

Could also just wrap headers in a `dict()` call at every call site to keep things simpler (trading some additional (probably negligible) memory copying for fewer lines + simpler code).

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -99,6 +89,30 @@ def _normalize_validate(validate):
         return validate
 
 
+def _get_property_name(name, field, context):
+    """
```
twosigmajab: (nit: rm leading \n from docstring (multiple occurrences))

tests/test_rebar.py
```diff
@@ -45,8 +46,14 @@ class FooListSchema(m.Schema):
...
 
+if MARSHMALLOW_MAJOR_VERSION == 2:
+    _name_field = m.fields.String(load_from='x-name', required=True)
+else:
+    _name_field = m.fields.String(data_key='x-name', required=True)
+
+
 class HeadersSchema(m.Schema):
-    name = m.fields.String(load_from='x-name', required=True)
+    name = _name_field
```
twosigmajab: since _name_field isn't used elsewhere, you could keep this a 1-liner with something like
```python
name = m.fields.String(required=True, **{('load_from' if MALLOWV2 else 'data_key'): 'x-name'})
```

If this load_from/data_key difference is needed elsewhere maybe `DATA_KEY = 'load_from' if MV2 else 'data_key'` is the right abstraction, so you could do ` m.fields.String(**{DATA_KEY: 'x-name'})` anywhere it's needed.

flask_rebar/request_utils.py
barakalon: This is a really good thought, but there's a couple reasons I think why we should leave as is:

> Doesn't this already exist in Python 3? https://docs.python.org/3/library/types.html#types.MappingProxyType (and for supporting python 2 as well there's https://pypi.org/project/dictproxyhack/)

Unfortunately, EnvironHeaders.__iter__ yields tuples of (key, value). We need it to return keys to look like a `dict`. So this won't work: https://github.com/eevee/dictproxyhack/blob/master/dictproxyhack.py#L138

> Another one-line approach (which is also natively supported in Python 2) could be to call Mapping.register(werkzeug.datastructures.Headers), making use of https://docs.python.org/3/library/abc.html#abc.ABCMeta.register

I'd rather avoid mucking with this, for reasons you mention.

> Could also just wrap headers in a dict() call at every call site to keep things simpler (trading some additional (probably negligible) memory copying for fewer lines + simpler code).

Another good idea, but headers are case insensitive, which `EnvironHeaders` handles for us. Calling `dict` won'y quite cut it, since it will make the keys case sensitive.

flask_rebar/swagger_generation/marshmallow_to_swagger.py
barakalon: Haha I've always preferred the leading \n
I think this is one of those things where whichever one you're used to just _feels_ better.
Maybe we should consider adding something like https://github.com/ambv/black here, so we don't have to debate such things?

flask_rebar/request_utils.py
barakalon: Haha gdammit.

barakalon:
> the general approach here seems to be to provide ad hoc compatibility conversion by scattering in-place schema mutations throughout the code, which can be more brittle than providing this via a dedicated, higher-level compatibility layer, which you could imagine releasing as an independent library that wouldn't be tied to this codebase.

I totally agree that this is brittle. It definitely felt brittle when I was throwing it together.

Regarding a higher-level compatibility layer in a separate package - I'd push back on it being separately package-able. If defining an interface that everyone wants for marshmallow was easy, marshmallow would've already done it and wouldn't need any breaking changes. 

Coming up with an interface that `flask-rebar` specifically needs seems more realistic. I was trying avoid a really heavy middle layer, since we depend on marshmallow internals so heavily. But let me see what I can do...

barakalon:
@twosigmajab This is starting to feel a little cleaner to me

flask_rebar/validation.py
```diff
@@ -89,11 +93,11 @@ def disallow_extra_fields(self, processed_data, original_data):
             raise ValidationError(message=messages.unsupported_fields(unsupported_fields))
 
 
-class RequestSchema(DisallowExtraFieldsMixin, Schema):
-    """
-
-    """
-    pass
+# Marshmallow version 3 starts "disallowing" extra fields by default this by default
```
twosigmajab: typo: ~this by default~

setup.py
```diff
@@ -15,8 +15,8 @@
         packages=find_packages(exclude=('test*', 'examples')),
         include_package_data=True,
         install_requires=[
-            'Flask>=0.10',
-            'marshmallow>=2.13',
+            'Flask>=0.10,<1',
+            'marshmallow>=2.13,<4',
```
twosigmajab: 👍

setup.py
```diff
@@ -15,8 +15,8 @@
         packages=find_packages(exclude=('test*', 'examples')),
         include_package_data=True,
         install_requires=[
-            'Flask>=0.10',
-            'marshmallow>=2.13',
+            'Flask>=0.10,<1',
```
twosigmajab: Typo: This should be pinning Flask to `<2`, not `<1`. Compatibility with Flask 1 is tested for in the .travis.yml above.

flask_rebar/compat.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def dump(schema, data):
+        schema.strict = True
+        data, _ = schema.dump(data)
+        return data
+
+    def exclude_unknown_fields(schema):
+        return schema
+
+else:
```
twosigmajab: This PR is currently getting rid of the runtime compatibility check on the marshmallow version (previously it was logging a warning if found to be outside expectation). 

If code throughout this package is going to branch off `if MALLOW_V2` and then only have an `else` clause that *assumes* V3, is it worth keeping a marshmallow version compatibility check at runtime?

flask_rebar/compat.py
```diff
@@ -0,0 +1,60 @@
+import sys
+from distutils.version import StrictVersion
+
+import marshmallow
+
+
+PY2 = int(sys.version_info[0]) == 2
+MARSHMALLOW_VERSION = StrictVersion(marshmallow.__version__).version
+MARSHMALLOW_MAJOR_VERSION = MARSHMALLOW_VERSION[0]
```
twosigmajab: It looks like everywhere this is used, it is used like `MARSHMALLOW_MAJOR_VERSION == 2`. That argues for factoring this value out into its own constant, e.g. `MALLOW_V2 = (MARSHMALLOW_MAJOR_VERSION == 2)`.

flask_rebar/compat.py
```diff
@@ -0,0 +1,60 @@
+import sys
+from distutils.version import StrictVersion
+
+import marshmallow
+
+
+PY2 = int(sys.version_info[0]) == 2
+MARSHMALLOW_VERSION = StrictVersion(marshmallow.__version__).version
```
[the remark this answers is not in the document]
barakalon: Ah I was unaware of that one. I'll switch

setup.py
barakalon: Whoops, good catch

flask_rebar/compat.py
barakalon: Now that we've specified the dependency properly in `setup.py`, tools like `pip` should give the warning for us. In the case where someone install marshmallow>=4 despite this warning, we should give our best effort support.

So I think the `else` is fine for now 🤷‍♂️

barakalon:
@brockhaywood any chance we can get this merged soon?

brockhaywood:
@twosigmajab you've been working on this with @barakalon closely. Can you update your review status?

tests/test_validation.py
```diff
@@ -166,20 +174,25 @@ class IntegerQuery(Schema):
...
-        _, errs = IntegerQuery().load(query)
-        self.assertEqual(errs, {
+
+        with self.assertRaises(ValidationError) as ctx:
+            compat.load(IntegerQuery(), query)
+
+        self.assertEqual(ctx.exception.messages, {
+            # Marshmallow's fields.List formats the dump errors differently
+            # than load :shrug:
             'foos': {1: ['Not a valid integer.']}
```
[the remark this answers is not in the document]
barakalon: Yeah I'm not sure what this is about either haha I'll remove it.

This repo doesn't have any sort of linting set up yet... so sure, 88 sounds great :) Even more is good with me too.

tests/test_errors.py
```diff
@@ -162,24 +163,40 @@ def test_json_body_parameter_validation(self):
...
+        if MARSHMALLOW_V2:
+            self.assertEqual(resp.json, {
+                'message': messages.body_validation_failed,
+                'errors': {'_general': 'Unexpected field: baz'}
+            })
+        else:
+            self.assertEqual(resp.json, {
+                'message': messages.body_validation_failed,
+                'errors': {'baz': 'Unknown field.'}
+            })
```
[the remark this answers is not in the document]
barakalon: You're right, but I'd prefer to pass on this for now. DRYing this up seems more complex than its worth (to me, at the moment)

barakalon:
@brockhaywood good to go 👍

twosigmajab:
Hi Flask-Rebar friends, looks like Marshmallow v3.0.0 final just came out. From https://marshmallow.readthedocs.io/en/stable/changelog.html it looks like there were several backwards-incompatible changes added to the Marshmallow v3 line *after* this PR landed. Now that v3 is final, is it worth double-checking that changelog to see if Rebar needs any other changes to maintain compatible behavior with both v2 and v3?

Sytten:
@barakalon @RookieRick Can we check that again? I will do a test today, but I fear the final release of marshmallow yesterday might have broken our package. Maybe push a small patch for now to limit the version of marshmallow to 2 (its currently at <4) until we figure it out.

RookieRick:
Eek, will take a look asap. On my way to vet appt for my dog so will be a little later this morning but yeah if there are issues I agree a quick patch to fix pin of marshmallow sounds like a great idea

RookieRick:
Ugh "later this morning"...  😂 ... so yeah looks like we also have a test that passes with 2.19.0 but fails with 2.20 and a few more tests that fail with 3.0 so for moment will just pin to <=2.19

PR #49, opened 2018-12-04
Bump v2.0.0

brockhaywood:


brockhaywood:
@barakalon seems to me the marshmallow 3 updates are worthy of a major bump? thoughts?

barakalon:
There _shouldn't_ be any backwards incompatible changes (let me know if you're aware of one?), so I would assume a minor change.

brockhaywood:
no, i'm not aware of any. just based on marshmallow 3 being a major version bump, this could be. but, i'm ok with a minor bump since there is nothing backwards incompatible for us.

twosigmajab:
This is tricky because the dependency declaration already declared support for `marshmallow>=2` even though it only supported `marshmallow>=2,<3`. If it had declared that, then this PR would include the change `marshmallow>=2,<4`, and in so doing, would warrant a major version bump.

PR #51, opened 2018-12-06
Support setting response content-types for endpoints.

juliusiv:
WIP. Currently it only hooks up the registry to the swagger so you can define the response type for a specific endpoint. We probably want to only allow certain content types here because I think we need special ways to "dump" things based on what content type it's supposed to be.


juliusiv:
i'm punting on this for now, gonna close this

PR #53, opened 2019-01-07
Add 'https' to default schemes

twosigmajab:
Flask-Rebar currently sets `schemes` to `('http',)` by default. As a result, users serving their app over HTTPS get broken "Try it out" buttons in their Swagger UI out of the box (because for security reasons the browser will refuse to make the XHRs over HTTP). To get around this, users have to manually add code like the following:
```python
registry.swagger_generator.schemes = ('http', 'https')
```
Now they will be able to select "https" from the "schemes" dropdown in the Swagger UI and the "Try it out" buttons will work.

What do you think of defaulting to `('http', 'https')` out of the box?

### DIFF
--- a/flask_rebar/swagger_generation/swagger_generator.py
+++ b/flask_rebar/swagger_generation/swagger_generator.py
@@ -273,7 +273,7 @@ class SwaggerV2Generator(object):
     def __init__(
             self,
             host='swag.com',
-            schemes=('http',),
+            schemes=('http', 'https'),
             consumes=('application/json',),
             produces=('application/json',),
             version='1.0.0',


twosigmajab:
Thanks for the quick review, @barakalon, glad this makes sense to you! @juliusiv @brockhaywood LGTY?

By the way, I just opened #54 after finding https://github.com/swagger-api/swagger-ui/issues/4710#issuecomment-405448739 -- if I'm understanding that correctly, when you load the Swagger UI over HTTPS, and both "http" and "https" are in `schemes`, the Swagger UI should automatically make requests over HTTPS instead of HTTP, but in my testing it doesn't; I have to manually change the "schemes" dropdown from HTTP to HTTPS to get this to work. Curious if this works for you. So I figured maybe the version of Swagger UI that Flask-Rebar is bundling predates that auto-HTTPS feature, but then I couldn't tell what version is bundled.

PR #55, opened 2019-01-25
Support for tags

barakalon:
Resolves this issue: https://github.com/plangrid/flask-rebar/issues/52

@twosigmajab

PR #59, opened 2019-02-12
chg: Updated todo example to show tag usage

kazamatzuri:
Hi folks,

to familiarize myself with this I updated the todo example with the tag usage that was added in #55. 
Hope that's going to be helpful to someone. 

Comments? 
Thanks!
-Fab

### DIFF
--- a/examples/todo/todo.py
+++ b/examples/todo/todo.py
@@ -4,13 +4,38 @@ from marshmallow import fields, pre_dump, Schema
 from flask_rebar import Rebar, errors, HeaderApiKeyAuthenticator
 from flask_rebar.validation import RequestSchema, ResponseSchema
 
+from flask_rebar import Tag,SwaggerV2Generator
 
 rebar = Rebar()
-registry = rebar.create_handler_registry()
+
+# add a description of the tags we're using further down (will be output to the generated swagger)
+# a bit reduntant, given the simplicity of the code, but you get the idea :) 
+generator = SwaggerV2Generator(
+    tags=[
+        Tag(
+            name='create',
+            description='All operations relating creation of todo items.'
+        ),
+        Tag(
+            name='read',
+            description='All operations relating to reading of todo items.'
+        ),
+        Tag(
+            name='update',
+            description='All operations relating to updating of todo items.'
+        )
+    ]
+)
+
+
+registry = rebar.create_handler_registry(swagger_generator=generator)
 
 # Just a mock database, for demonstration purposes
 todo_id_sequence = 0
 todo_database = {}
+
+
+
 
 
 # Rebar relies heavily on Marshmallow.
@@ -57,7 +82,7 @@ class TodoListSchema(ResponseSchema):
     rule='/todos',
     method='POST',
     request_body_schema=CreateTodoSchema(),
-
+    tags=['create'],
     # This dictionary tells framer which schema to use for which response code.
     # This is a little ugly, but tremendously helpful for generating swagger.
     marshal_schema={
@@ -87,7 +112,7 @@ def create_todo():
     rule='/todos',
     method='GET',
     query_string_schema=GetTodoListSchema(),
-
+    tags=['read'],
     # If the value for this is not a dictionary, the response code is assumed
     # to be 200
     marshal_schema=TodoListSchema()
@@ -113,7 +138,8 @@ def get_todos():
     rule='/todos/<int:todo_id>',
     method='PATCH',
     marshal_schema=TodoResourceSchema(),
-    request_body_schema=UpdateTodoSchema()
+    request_body_schema=UpdateTodoSchema(),
+    tags=['update']
 )
 def update_todo(todo_id):
     global todo_database
--- a/examples/todo/todo.py
+++ b/examples/todo/todo.py
@@ -16,8 +16,7 @@ generator = SwaggerV2Generator(
     tags=[
         Tag(
             name='todo',
-            description='All operations to managing the todo list portion of \
-            the API'
+            description='Operations for managing TODO items.'
         ),
     ]
 )
--- a/examples/todo/todo.py
+++ b/examples/todo/todo.py
@@ -1,29 +1,24 @@
 from flask import Flask
 from marshmallow import fields, pre_dump, Schema
 
-from flask_rebar import Rebar, errors, HeaderApiKeyAuthenticator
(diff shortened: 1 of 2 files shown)


examples/todo/todo.py
```diff
@@ -4,15 +4,40 @@
 from flask_rebar import Rebar, errors, HeaderApiKeyAuthenticator
 from flask_rebar.validation import RequestSchema, ResponseSchema
 
+from flask_rebar import Tag,SwaggerV2Generator
```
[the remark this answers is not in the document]
kazamatzuri: no worries. I'll change it. thanks!

examples/todo/todo.py
```diff
@@ -4,15 +4,40 @@
...
 
 rebar = Rebar()
-registry = rebar.create_handler_registry()
+
+# add a description of the tags we're using further down (will be output to the generated swagger)
+# a bit reduntant, given the simplicity of the code, but you get the idea :) 
+generator = SwaggerV2Generator(
```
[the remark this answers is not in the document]
kazamatzuri: Hm, I hadn't thought about it that way, as you only need to pull in the generator to add the tag descriptions. Good point though, I'll update that as well.

examples/todo/todo.py
```diff
@@ -4,15 +4,40 @@
...
+    tags=[
+        Tag(
```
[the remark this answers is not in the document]
kazamatzuri: No need for self hating;) I was struggling with that a bit myself, I thought these would make more sense than just a dummy tag. But since I only wanted it to demonstrate how to use the tagging feature, I don't feel too strongly about that.

examples/todo/todo.py
```diff
@@ -4,15 +4,40 @@
...
+            name='create',
+            description='All operations relating creation of todo items.'
+        ),
+        Tag(
+            name='read',
+            description='All operations relating to reading of todo items.'
+        ),
+        Tag(
```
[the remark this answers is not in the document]
kazamatzuri: ok. I'll update this and all the above.

examples/todo/todo.py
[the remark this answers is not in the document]
kazamatzuri: If I do that and we're sticking with pep8 for now, the line becomes too long. the \ continuation looks rather ugly (imho). 
I think the output hasn't been updated in a while, there are couple of other changes that are note related to my doing (i think). https scheme got add ed and the producing mime type is fixed now.

examples/todo/todo.py
```diff
@@ -1,29 +1,24 @@
...
+# However, if you want more control over how the swagger is generated, you can 
+# provide your own. 
+# Here we've specified additional metadata for operation tags.
 generator = SwaggerV2Generator(
     tags=[
         Tag(
-            name='create',
-            description='All operations relating creation of todo items.'
+            name='todo',
+            description='All operations to managing the todo list portion of \
```
barakalon: nitpick: this is phrased a bit weird. Maybe: "Operations for managing TODO items"?

examples/todo/todo.py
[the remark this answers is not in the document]
barakalon: Ah, I think a common addition to vanilla pep8 is to extend the line length. 80 is a bit restrictive.

examples/todo/todo_output.md
```diff
@@ -10,7 +10,6 @@ $ curl -s -XGET http://127.0.0.1:5000/swagger
   ],
```
barakalon: @kazamatzuri Thanks for updating this!

barakalon:
Sweet 👍 !

examples/todo/todo.py
kazamatzuri: must have been the cat walking on the keyboard ;)

examples/todo/todo_output.md
kazamatzuri: np

examples/todo/todo.py
```diff
@@ -1,12 +1,28 @@
 from flask import Flask
 from marshmallow import fields, pre_dump, Schema
 
-from flask_rebar import Rebar, errors, HeaderApiKeyAuthenticator
+from flask_rebar import Rebar, errors, HeaderApiKeyAuthenticator, Tag, \
+    SwaggerV2Generator
```
twosigmajab: Check out https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

Would this all fit on one line using an 88-char max line length?

examples/todo/todo.py
```diff
@@ -134,9 +151,10 @@ def create_app(name):
     app = Flask(name)
 
     authenticator = HeaderApiKeyAuthenticator(header='X-MyApp-Key')
-    # The HeaderApiKeyAuthenticator does super simple authentication, designed for
-    # service-to-service authentication inside of a protected network, by looking for a
-    # shared secret in the specified header. Here we define what that shared secret is.
+    # The HeaderApiKeyAuthenticator does super simple authentication, designed 
+    # for service-to-service authentication inside of a protected network, by 
+    # looking for a shared secret in the specified header. Here we define what 
+    # that shared secret is.
```
twosigmajab: (ditto about bumping to 88)

examples/todo/todo.py
kazamatzuri: No.
Here's what black makes of the imports (with the default config):
`from flask import Flask
from marshmallow import fields, pre_dump

from flask_rebar import (
    Rebar,
    errors,
    HeaderApiKeyAuthenticator,
    Tag,
    SwaggerV2Generator,
)
from flask_rebar.validation import RequestSchema, ResponseSchema`

kazamatzuri: note: I will not add this change here, since it is a style migration that is independent of original scope of 'add tags to the todo example'. I'm happy to put this into a new PR if that is what we decide (see #60 )

examples/todo/todo.py
kazamatzuri: This becomes:
`    # The HeaderApiKeyAuthenticator does super simple authentication, designed for
    # service-to-service authentication inside of a protected network, by looking for a
    # shared secret in the specified header. Here we define what that shared secret is.
`
note: I will not add this change here, since it is a style migration that is independent of original scope of 'add tags to the todo example'. I'm happy to put this into a new PR if that is what we decide (see #60 )

PR #65, opened 2019-02-21
fix: Do not rethrow redirect errors

### LINKED ISSUE #24: _init_error_handling() breaks Flask's handling of RequestRedirect when not in debug mode

juliusiv:
Fixes #24, see that issue for a discussion.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@ Changelog
 =========
 
 
+(unreleased)
+------------
+
+Fix
+~~~
+- Do not rethrow redirect errors. [Julius Alexander]
+
+
 v1.4.1 (2019-02-19)
 -------------------
 
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -618,15 +618,12 @@ class Rebar(object):
 
         @app.errorhandler(REDIRECT_ERROR)
         def handle_request_redirect_error(error):
-            if current_app.debug:
-                raise error
-            else:
-                return self._create_json_error_response(
-                    message=error.name,
-                    http_status_code=error.code,
-                    additional_data={"new_url": error.new_url},
-                    headers={"Location": error.new_url}
-                )
+            return self._create_json_error_response(
+                message=error.name,
+                http_status_code=error.code,
+                additional_data={"new_url": error.new_url},
+                headers={"Location": error.new_url}
+            )
 
         @app.errorhandler(Exception)
         def handle_generic_error(error):
--- a/tests/test_rebar.py
+++ b/tests/test_rebar.py
@@ -566,5 +566,7 @@ class RebarTest(unittest.TestCase):
         self.assertTrue(resp.headers['Location'].endswith('/with_trailing_slash/'))
 
         app.debug = True
-        with self.assertRaises(RequestRedirect):
-            app.test_client().get(path='/with_trailing_slash')
+        resp = app.test_client().get(path='/with_trailing_slash')
+        self.assertEqual(resp.status_code, 301)
+        self.assertEqual(resp.content_type, 'application/json')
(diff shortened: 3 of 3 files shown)


twosigmajab:
Hi Flask-Rebar maintainers, would you be interested in making a new release with this fix? Seeing users hitting this and scratching their heads.
Thanks again for Flask-Rebar, it's fantastic! /cc @RookieRick, @juliusiv, @brockhaywood et al.

RookieRick:
> Hi Flask-Rebar maintainers, would you be interested in making a new release with this fix? Seeing users hitting this and scratching their heads.
> Thanks again for Flask-Rebar, it's fantastic! /cc @RookieRick, @juliusiv, @brockhaywood et al.

Gut instinct answer is "yes" but as I'm still relatively new to PlanGrid and associated ecosystem let me circle the wagons internally and make sure I'm not going to break anything ;-)

PR #68, opened 2019-02-24
chg: Enforce black on PR's

juliusiv:


juliusiv:
@kazamatzuri Alright, I got Travis working. I have it only checking if the code is `black` formatted if we're using python 3.6 because it won't run on any version less than that. It's a bit weird to me that we're running it like this because we're dependent on the python version in the build of the version we're checking compatibility for but I wasn't sure how else to do it. As long as it gets run once that's cool with me.

@RookieRick mind taking a peek at this?

PR #73, opened 2019-03-22
fix: Werkzeug 0.14->0.15 introduced some breaking changes in redirects

RookieRick:


tests/test_errors.py
```diff
@@ -65,6 +65,10 @@ def test_default_404_errors_are_formatted_correctly(self):
             "the URL manually please check your spelling and try again."
         )
         resp = self.app.test_client().get("/nonexistent")
+        # Update of Werkzeug 0.14 -> 0.15 broke this test with a change from double to single space in message.
+        # To avoid this sort of silliness in the future, we'll normalize to single-space:
+        msg = msg.replace("  ", " ")
+        resp.json["message"] = resp.json["message"].replace("  ", " ")
         self.assertEqual(resp.status_code, 404)
         self.assertEqual(resp.content_type, "application/json")
         self.assertEqual(resp.json, {"message": msg})
```
twosigmajab: Would it be sufficient to merely check for the 404 status here, to be less breakage-prone if the text changes again in the future? Or at most just `assert "not found" in msg.lower()` or some such?

RookieRick: Good question - this whole series of tests has names like "test_default_{code}_errors_are_formatted_correctly" and all check the returned message as well as status code and content type, so I just went on the assumption that somebody somewhere cared about that level of detail.  All that said, some historical perspective as to why these particular tests exist and whether we really need to verify the wording of the error message might be in order :) (cc @barakalon @juliusiv who it looks like were primary contributors to this file originally)

barakalon: A lot of flask-rebar was extracted from PlanGrid's external facing API at the time, which had very thorough tests like this.

I think it makes sense to check the message for `400` errors, since there's a chance that a _different_ 400 error was thrown than the one the test is expecting.

But I agree, verifying the message for `404` seems excessive. Maybe just check that `"message" in resp.json`?

PR #78, opened 2019-04-09
Add swagger mapping for oid_string

aebrow4:
Add a mapping for a Flask converter of type oid_string to Swagger's
string type.

### DIFF
--- a/flask_rebar/swagger_generation/swagger_generator.py
+++ b/flask_rebar/swagger_generation/swagger_generator.py
@@ -376,6 +376,7 @@ class SwaggerV2Generator(object):
         self.flask_converters_to_swagger_types = {
             "uuid": sw.string,
             "uuid_string": sw.string,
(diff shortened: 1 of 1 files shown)


aebrow4:
OK, thanks Barak - makes sense knowing that uuid_string isn't meant to be there. flask-toolbox doesn't do any mapping of converters-swagger mapping, but it can.

airstandley:
I agree with @barakalon that the definition of the OID converts belongs in flask-toolbox as part of the bootstrap, but it might not be a bad idea to add better support for registering custom converts in rebar. 

Rebar is supposed to bridge the gap between flask and swagger. It seems a bit clunky to have to register custom converters with flask, and then register the convert with the SwaggerGenerator. I'd much rather be able to make a single call to rebar.

barakalon:
@airstandley I like that idea.

Issue #18, opened 2018-10-10
Support OpenAPI Spec version 3?


twosigmajab:
It looks like flask-rebar provides a `SwaggerV2Generator` but not a `SwaggerV3Generator` (or similar). OpenAPI Spec v3 was released last year (see https://github.com/OAI/OpenAPI-Specification/releases). Is there any interest in adding support for this to flask-rebar?

barakalon:
There's definitely _interest_.

OpenAPI v3 is still relatively new, and community tooling hasn't quite caught up yet (e.g. https://github.com/swagger-api/swagger-codegen). PlanGrid chose to stick with v2 for this reason, and there hasn't been a call to create SwaggerV3Generator yet - this would be that first call!

Is support for v3 a requirement for you? Or are you just curious about goals?

Also... I can take a swing at SwaggerV3Generator, but it might take me some time. Any interest in contributing?

twosigmajab:
We are currently using version 2, but in the coming months intend to evaluate updating to 3 along with looking at https://github.com/OpenAPITools/openapi-generator, the Swagger Codegen fork that also supports OAI 3. Good to see that contributing this to Flask-Rebar would be considered, will definitely keep that in mind.

juliusiv:
+1 for support for this. Swagger2's lack of allowing null values would be nice. We've noticed we need to dumb down schemas a bit because of that.

barakalon:
Yeah, this is inevitable and increasingly important.

I've been meaning to hop on this, but haven't had time.

If anyone else wants to handle this change... 😬

joeb1415:
Any updates on this issue? I am considering working on this as a hackweek project next week, curious if anyone has put any more thought or time into it in the past few months. 

In addition to the null values Julius mentioned, I am excited to add support for OneOf as well, as we discussed a while back here: https://github.com/plangrid/flask-rebar/pull/48#issuecomment-443805956.

barakalon:
I haven’t put in any additional thought/time. 

This would be a fantastic contribution! I tried my best to write the existing code in preparation for this, but who knows what will happen. 

I’m happy to lend some initial thoughts, maybe a walkthrough the relevant code, if you’re interested.

Sytten:
We can close this one @RookieRick

RookieRick:
🎉

PR #80, opened 2019-04-15
Add OpenAPI 3 Support

barakalon:
Alrighty, here's a stab at support for OpenAPI 3.0 (https://github.com/plangrid/flask-rebar/issues/18)

I still have to update documentation and add some tests, but I wanted to get this up sooner rather than later so we can talk about certain interface decisions.


barakalon:
How do y'all think we should approach the API endpoints for swagger?

Should we have a different path for 2.0 and 3.0?
Should we server 2.0 OR 3.0, and require developers to specify which one (defaulting to 2.0, for backwards compatibility)
Something else?

Sytten:
I think the second approach (2.0 OR 3.0 with default to 2.0) is the right approach!

RookieRick:
On the 2.0 vs 3.0 question, I agree allowing developer to specify 2.0 or 3.0 with 2.0 default is ideal - bonus points if we can make it intuitively obvious that this is happening so a brand new Flask-Rebar developer is likely to see it and go "Oh! I can do 3.0?"

barakalon:
> bonus points if we can make it intuitively obvious that this is happening so a brand new Flask-Rebar developer is likely to see it and go "Oh! I can do 3.0?"

That's a good thought. Aside from adding a section in the documentation about swagger generation, here's the current story:

When instantiating a `Rebar` instance and a handler registry, one hint is that `create_handler_registry` has a `swagger_generator` parameter. Depending on your workflow/tools, you might see this. Here's PyCharm: 

image

Both `SwaggerV2Generator` and `SwaggerV3Generator` are exported from the root package, so again, depending on your tools, you might see them both. Here's PyCharm:

image

Anything else we can do?

RookieRick:
> Anything else we can do?

I'd say the docs + parameters hints are sufficient hand-holding.  If somebody's using Vi, I'll assume they know how to rtfm! 🤣

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -0,0 +1,322 @@
...
+        schemes=None,
+        consumes=None,
+        produces=None,
+        sort_keys=True,
+    ):
+        """Generates a Swagger specification from a Rebar instance.
+
+        :param flask_rebar.rebar.HandlerRegistry registry:
+        :param str host: Adds this host as a Server Object for the service
+        :param schemes: IGNORED, but kept for compatibility with SwaggerV2Generator
```
joeb1415: Do we need compatibility? If I change my app from `SwaggerV2Generator.generate()` to `SwaggerV3Generator.generate()`, I would expect that the params might change.

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -0,0 +1,322 @@
...
+
+                request_body = None
+
+                if d.request_body_schema:
+                    schema = d.request_body_schema
+
+                    request_body = {
+                        sw.required: True,
+                        sw.content: {
+                            "application/json": {
```
joeb1415: Should this be configurable via `consumes`? Or are we just saying JSON only?

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -0,0 +1,322 @@
...
+                if d.tags:
+                    path_definition[method_lower][sw.tags] = d.tags
+
+        return path_definitions
+
+    def _get_response_definition(self, schema):
+        return {
+            sw.description: get_response_description(schema),
+            sw.content: {
+                "application/json": {sw.schema: get_ref_schema(self._ref_base, schema)}
```
joeb1415: Same question re: configurable via `produces`?

flask_rebar/swagger_generation/swagger_generator_v2.py
```diff
@@ -0,0 +1,317 @@
...
+            sw.schemes: list(schemes or self.schemes),
+            sw.consumes: list(consumes or self.consumes),
+            sw.produces: list(produces or self.produces),
+            sw.security_definitions: security_definitions,
+            sw.paths: paths,
+            sw.definitions: definitions,
+        }
+
+        if default_authenticator:
+            swagger[
```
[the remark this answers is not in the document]
barakalon: I thought `black` was supposed to solve this 😂 
I didn't format it like this - `black` did.

flask_rebar/swagger_generation/swagger_generator_v3.py
barakalon: I think the best thing to do is add proper content negotiation to flask-rebar (@juliusiv had a PR about this a while ago: https://github.com/plangrid/flask-rebar/pull/51)
We could also allow "consumes" and "produces" to be set, and use this value accordingly, but that will be shoving the Swagger v2 interface into this one.

barakalon: I was hoping to solve that problem after this PR, since this one got hefty enough.

Do you think we need to suppose the whole "application/json+plangrid" thing before releasing this?

flask_rebar/swagger_generation/swagger_generator_v2.py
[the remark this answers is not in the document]
RookieRick: Marking this one as "resolved" but agreed all around - ewww! LOL

flask_rebar/compat.py
```diff
@@ -1,15 +1,21 @@
...
+
+if int(sys.version_info[0]) == 2:
+    Mapping = collections.Mapping
+else:
+    Mapping = collections.abc.Mapping
 
 
-if PY2:
-    from collections import Mapping
+if sys.version_info >= (3, 4):
```
RookieRick: I don't see ABC referenced anywhere - is this (still, or soon-to-be) needed? (Or some mystical Python trickery? :D )

Sytten: IMO, in modern Python, any base class should inherit from ABC.

flask_rebar/swagger_generation/swagger_generator_v3.py
RookieRick: Personally I'm OK with leaving this hard-coded as-is and opening an issue to add proper content-negotiation (which, if it's required for something to move from v2 to v3, should be a great motivator ;))
Alongside that, I'd also suggest we remove those extra params noted above.

flask_rebar/swagger_generation/swagger_generator_v3.py
RookieRick: I'd agree (that we probably don't need to include these), particularly given the notes below about consumes/produces.  I could see an ignored param being confusing for somebody that's starting with V3

flask_rebar/compat.py
RookieRick: Oh, GitHub outsmarted me - the reference was hidden behind a "Load Diff" for a large file.  Carry on... :)

flask_rebar/swagger_generation/swagger_generator_v3.py
RookieRick: Assuming I haven't abused GitHub's forking and PR too badly for it to do what I intended, pushed a PR to Barak's fork which I think should address this: https://github.com/barakalon/flask-rebar/pull/1

barakalon: This should be addressed now

RookieRick:
Playing around with this a little and it looks like https://github.com/barakalon/flask-rebar/blob/openapi3-support/flask_rebar/swagger_generation/generator_utils.py#L136-L137 is choking on "anyOf" when I have a schema that defines something as required=False, missing=None.

barakalon:
@RookieRick Good catch!

Had to brush off my DFS skills for this one. Should be fixed now, and I think the code is actually cleaner than before.

RookieRick:
> @RookieRick Good catch!
> 
> Had to brush off my DFS skills for this one. Should be fixed now, and I think the code is actually cleaner than before.

Awesome - Sorry it took so long for me to get back to it - was gonna do over the weekend but then the weekend got in the way :D  Ran some quick tests with the same input as before and things are looking good -- I'll poke around a bit more and then try and release today 🎉

flask_rebar/swagger_generation/swagger_generator_v3.py
RookieRick: Per follow-on conversations with Joe, we can punt this to later if it becomes needed

PR #81, opened 2019-04-15
sort required array

BrandonWeng:
I believe ` obj.fields.items():` does not consistently return objects in the same order. When committing the generated `swagger.json` file, it causes some redundant diffs in the `required` fields array 

e.g: https://github.com/plangrid/informant/pull/419

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -199,7 +199,9 @@ class SchemaConverter(MarshmallowConverter):
                 prop = compat.get_data_key(field)
                 required.append(prop)
 
-        return sorted(required) if required else UNSET
+        if required and not obj.ordered:
+            required = sorted(required)
+        return required if required else UNSET
 
     @sets_swagger_attr(sw.description)
     def get_description(self, obj, context):
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -199,7 +199,7 @@ class SchemaConverter(MarshmallowConverter):
                 prop = compat.get_data_key(field)
                 required.append(prop)
 
-        return required if required else UNSET
+        return sorted(required) if required else UNSET
 
     @sets_swagger_attr(sw.description)
     def get_description(self, obj, context):
--- a/tests/swagger_generation/test_marshmallow_to_swagger.py
+++ b/tests/swagger_generation/test_marshmallow_to_swagger.py
@@ -176,7 +176,7 @@ class TestConverterRegistry(unittest.TestCase):
 
         schema = Foo()
         json_schema = registry.convert(schema)
-        print(json_schema)
+
         self.assertEqual(
             json_schema,
             {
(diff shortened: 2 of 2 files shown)


BrandonWeng:
Thanks for the reminder, updated comment to be a little more open source friendly  👍

airstandley:
Marshmallow SchemaMeta has the [option to make the fields be ordered](https://marshmallow.readthedocs.io/en/3.0/quickstart.html#ordering-output). 
Any reason we can't just leverage that functionality?

(Side note: Technically not an issue for Python 3.6+ since they changed the underlying hash table implementation and then made the decision to guarantee that the ordering would be determinant from then on. )

BrandonWeng:
If a service wants to take advantage of the functionality, would we have to explicitly state `ordered = True` for every schema? Or maybe have a base schema to inherit from? Sorry, I'm not too familiar with Marshmallow and the client generation details

Agreed on the side note. Do we have plans to bump our python version in flask-rebar?

airstandley:
Flask Rebar already defines `RequestSchema` and `ResponseSchema` base schema objects in `validation.py`. Service schemas should already be deriving from these, so we could set ordering= True on them. I can't see it being a breaking change, but it would definitely require a second review.

> Do we have plans to bump our python version in flask-rebar?

flask-rebar works with Python 3.6 already.

airstandley:
Looping in Rick

barakalon:
The tests are the source of truth for all versions of Python that we officially support: https://github.com/plangrid/flask-rebar/blob/master/.travis.yml

Probably a good idea to add 3.7 in there.

Sorting the output of required seems practical, since the marshmallow to swagger conversion doesn't actually care if the marshmallow schema is a `RequestSchema` or a `ResponseSchema`, and I don't think it should care. 

Maybe do both?

Sytten:
It already works with Python 3.7, we use it daily.
It would be interesting to not break the functionality if you don't use the provided schema. On our side, we use https://github.com/SVilgelm/marshmallow-objects instead.

RookieRick:
OK I think if I've followed everything correctly here, this makes sense to do it as Brandon did in order to not break functionality for anybody using their own schemas (not that it would "break" necessarily in any event, but "break" in terms of producing inconsistent swagger file).  
Sytten especially, if I'm wrong on that and this causes any issues with your use of marshmallow-objects please let me know as I definitely don't want to remove support for that.
Will create a sep issue to update our tests - nice catch Barak :)

airstandley:
I'd prefer it is we confirmed the fields were not already ordered before doing the sort. It's kind of count intuitive behaviour for the generated swagger definition to re-order the required fields when you've explicitly ordered them.

barakalon:
@airstandley ah, I see what you're saying. Good point.

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -199,7 +199,7 @@ def get_required(self, obj, context):
                 prop = compat.get_data_key(field)
                 required.append(prop)
 
-        return required if required else UNSET
+        return sorted(required) if required else UNSET
```
airstandley:
```suggestion
        if required and not obj.ordered:
            required = sorted(required)
        return required if required else UNSET
```

tests/swagger_generation/test_marshmallow_to_swagger.py
```diff
@@ -189,6 +189,7 @@ class Foo(m.Schema):
 
     def test_required(self):
         class Foo(m.Schema):
+            b = m.fields.Integer(required=True)
```
joeb1415: could add a `test_ordered` as well for that case

airstandley: Good idea.

Issue #82, opened 2019-04-16
Add Python 3.7 to test suite


RookieRick:

RookieRick:
(Also some other similar flavored housekeeping like updates to Black etc)

RookieRick:
Addressed by #83

PR #83, opened 2019-04-17
@minor: updating test config and tweaking some docs

RookieRick:
Addresses #82


RookieRick:
So it seems it's not as easy as just adding "3.7" - lol SIIIIIGH .   ref https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905

PR #93, opened 2019-05-09
Return empty object not empty string, if an empty non-null object response is specified

joeb1415:


flask_rebar/request_utils.py
```diff
@@ -52,22 +52,19 @@ def __getitem__(self, key):
...
     :param int status_code: HTTP status code to use in the response
+    :param dict headers: Additional headers to attach to the response
     :rtype: flask.Response
     """
-    resp = jsonify(data)
-
-    if not get_json_from_resp(resp=resp):
-        resp.data = ""
+    if data is None:
+        resp = Response(mimetype="application/json")
```
joeb1415: I don't love specifying this here, but if we don't, then we get `text/plain` from [werkzeug](https://github.com/pallets/werkzeug/blob/0.15.2/src/werkzeug/wrappers/base_response.py#L139), and jsonify barfs on passing None as args on old versions of [Flask](https://github.com/pallets/flask/blob/0.10/flask/json.py#L222).

airstandley:
@joeb1415 Can you point me at the issue this addresses?

RookieRick:
> @joeb1415 Can you point me at the issue this addresses?

@airstandley This was regarding Barak's observation at https://github.com/plangrid/flask-rebar/pull/85#issuecomment-490659721

flask_rebar/request_utils.py
airstandley: This will enforcing a content-type on schema-less empty responses.

RookieRick: A couple quick thoughts before I run out of brain-power for the day 🤣  (@joeb1415 @airstandley let's chat tomorrow morning to sync up - @barakalon would of course love to get your continued input here and/or we can get an email chain going.  @sytten I think you might have the closest thing to a known real-world use-case here too.  And anybody else that's following along of course! :) )

1. Agreed, this will enforce a content-type on schema-less empty responses
2. Presumably we could soften this a bit by making it just a default (i.e., if `headers` includes a `Content-Type` header, parrot that, else use "application/json")
2a.  Maybe instead of hard-coded "application/json" here we could/should future-proof a little by defining a `default_content_type` (or if we want to be consistent with Werkzeug and Flask terminology we could call it `default_mimetype`) which could be used here and elsewhere where we have similar situation (e.g, in the new `swagger_generator_v3` there's currently hard-coded "application/json" as well)
3. Assuming we decide that is sufficient to address the issue at-hand, it does cause a change in behavior..  Prior, empty schema-less responses returned with a Content-Type of "text/html".  So we need to decide if we consider that previous "text/html" behavior a "bug" in which case we're fixing it, or if this would be a breaking change unless we go through extra contortions to match that old behavior.

joeb1415: Good summary Rick, thanks! 

1. Yes, this enforces a content-type on empty responses, but this is already the case from werkzeug, we're just changing it to be json instead of html. FWIW, [HTTP spec](https://tools.ietf.org/html/rfc2616#section-7.2.1) isn't strict on this, saying only that we should include the header if there is content, but not enforcing it and not specifying if it should/should not be present if there is no content. I could go either way on this one.

2. I like the idea to define a `default_content_type`, I guess this would be set when instantiating the registry? Perhaps, if we have the default, then we can set it even on No Content responses, but if it is not set, then we could omit that response header.

3. I think this is fixing a bug, certainly `text/html` was a non-sensical default content type to be including before.

joeb1415: Having trouble finding a clean way to remove the header. Even if we remove it from our constructed response, flask seems to run it thru a [response wrapper](https://github.com/pallets/werkzeug/blob/0.15.2/src/werkzeug/test.py#L1021) again, which will add the Content-Type header back in again, before delivering it.

joeb1415:
I've added a `default_mimetype` to the registry. LMK what folks think. 

Still doesn't address the "no Content-Type if No Content" issue, but again, that seems to be thoroughly baked in to Flask, and not obviously a requirement so I'm not sure it's something we need to fix anyway.

flask_rebar/request_utils.py
barakalon: @joeb1415 I think its customary to leave the header, and I agree with your point - `text/html` is just as arbitrary as `application/json`. So I don't think its a big deal to got with `application/json`.

With that, +1 for for (2) in @RookieRick 's comment above.

flask_rebar/request_utils.py
```diff
@@ -52,22 +52,20 @@ def __getitem__(self, key):
...
     :rtype: flask.Response
     """
-    resp = jsonify(data)
-
-    if not get_json_from_resp(resp=resp):
-        resp.data = ""
+    if data is None:
+        resp = Response(mimetype=mimetype)
+    else:
+        resp = jsonify(data)
```
barakalon: I think we should use `mimetype` in all cases, not just if `data is None`. As a developer, if I set the `mimetype` on a handler, I would expect it to return that mimetype.

RookieRick:
> I dig it! One small comment.
> 
> @RookieRick I think we may need to start considering [pre-releases](https://www.python.org/dev/peps/pep-0440/#pre-releases) and/or [developmental releases](https://www.python.org/dev/peps/pep-0440/#developmental-releases). We've now introduced a parameter and then immediately taken it away, which is technically a major change if we released it with a final version.

Agreed - I have been pondering this in back of mind during this incident; will take a look at those practices and see how we can tweak our workflow to continue facilitating rapid development without borking things for our hopefully-growing user base :)

flask_rebar/request_utils.py
airstandley: For the record I don't disagree with the point that "application/json" is as valid/arbitrary as "text/html" for an empty response. (And the RFC standard is not at all helpful on the matter.)

However if apps were breaking from getting "text/html" instead of "application/json" as the content-type header, then I think we have to assume changing "text/html" to "application/json" may also be a breaking change.

I definitely like this default_mimetype approach. Most people will ignore it and get a sensible default, but if anyone needs the content-type for empty responses to remain "text/hml" they can have that.

joeb1415:
Updated to respect the user-provided `mimetype` in all cases if specified.

I agree with a move to pre-releases or developmental releases. I was cringing making a breaking change here to take out that parameter.

flask_rebar/request_utils.py
RookieRick: Rerunning and stepping through a couple of tests on the latest, this should now even get us back to where we were if they don't specify anything (i.e., if nothing is specified for default, you'll get text/html for empty string).  🎉 

Planning to go ahead and push this out later today as another patch release on the assumption that our breaking change was short-lived enough (and anybody impacted by it would almost certainly have been following this conversation :) ) with the commitment to move to pre-release and/or development releases going forward.  Given the horse is already out of the barn and we're unbreaking, that seems preferable vs doing a major release for a breaking change that has been reverted...  Not what I'd call "ideal" but "good enough" :)

If that gives anybody excessive heartburn, let me know - thanks all!!

twosigmajab: worth pulling out this string literal into a meaningfully-named constant (`DEFAULT_MIMETYPE` or some such)?

twosigmajab: (sorry for the noise, realized I was looking at an outdated diff, please ignore. deleting the previous comment and this one now.)

PR #99, opened 2019-06-05
fixes a bug where http 400s are returned as http 500s

brockhaywood:
this is for a case where a werkzeug badrequest exception is raised
before the rebar handlers get invoked. this was causing the
default rebar exception handler to run, thus returning a 500

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -656,6 +656,7 @@ class Rebar(object):
                 additional_data=error.additional_data,
             )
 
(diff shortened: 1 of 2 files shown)


airstandley:
@brockhaywood So was the ultimate cause of the 500's just the generic_error handler being invoked?

airstandley:
Actually, thinking about it shouldn't be just be funneling all `wzerkzeug.exceptions.HTTPException` exceptions to the wzerkgeug handler?

PR #101, opened 2019-06-07
new: Added support for deprecating function parameters 

### LINKED ISSUE #89: Rename marshal_schema to response_body_schema
To better match `request_body_schema`. (First deprecate but continue to support `marshal_schema` as an alias of `response_body_schema` until Rebar v2.)

RookieRick:
..and used it to rename marshal_schema to response_body_schema.

~Note:  it is still referred to as "marshal_schema" in the `PathDefinition` named tuple to avoid potential breaking change just in case anybody is using that tuple definition for some (unintended by us at least?) purpose.  Could be a candidate for a hard rename of that in v2.0 when we deprecate `marshal_schema` altogether..~

Design decision on which I could easily be swayed..  I opted for making `FutureWarning` the default (so as to trigger stderr warnings by default) vs `DeprecationWarning` (ignored by default) when a deprecated parameter name is encountered.

Resolves #89


flask_rebar/rebar.py
```diff
@@ -56,6 +57,7 @@
 
 # To catch redirection exceptions, app.errorhandler expects 301 in versions
 # below 0.11.0 but the exception itself in versions greater than 0.11.0.
+# NOTE: as of Flask 1.0.3, redirects are no longer bubbled up as exceptions
```
RookieRick: Not related in any way to current changes, but figured it wouldn't hurt to add some clarity re: recent shenanigans

twosigmajab:
> Note: it is still referred to as "marshal_schema" in the PathDefinition named tuple to avoid potential breaking change...

You can subclass the namedtuple class to add a `@property` that serves as an alias. So instead of:
```python
PathDefinition = namedtuple("PathDefinition", ...)
```
You can do something like:
```python
class PathDefinition(namedtuple("_PathDefinition", "new_field_name ...")):
    @property
    def old_field_name(self):
        # log a deprecation warning here
        return self.new_field_name
```
What do you think?

Thanks for making this change, @RookieRick! I've been teaching a bunch of folks how to use Flask-Rebar and I think this change will make the API that much easier to learn and remember. ;)

RookieRick:
Interesting idea @twosigmajab and thanks for the head-start on it! 😄  I'll play around with that a bit as well as it would be nice if we could cleanly deprecate it in that case too

RookieRick:
> Interesting idea @twosigmajab and thanks for the head-start on it! 😄 I'll play around with that a bit as well as it would be nice if we could cleanly deprecate it in that case too

Insert maniacal mad-scientist cackling here..
It's aliiiiive - and the previously defined deprecated_parameter stuff bolted on nicely to `__new__` in our new PathDefinition class.  huzzah!

flask_rebar/rebar.py
```diff
@@ -33,29 +34,12 @@
 from flask_rebar.request_utils import get_query_string_params_or_400
 from flask_rebar.swagger_generation import SwaggerV2Generator
 from flask_rebar.swagger_ui import create_swagger_ui_blueprint
-
```
RookieRick: migrated to a subclass so we could rename `marshal_schema` to `response_body_schema` and add a to-be-deprecated `marshal_schema` property for backward compatibility.  Moved elsewhere in the file to sit with other class defs

flask_rebar/rebar.py
```diff
@@ -33,29 +34,12 @@
...
-# Metadata about a declared handler function. This can be used to both
-# declare the flask routing and to autogenerate swagger.
-PathDefinition = namedtuple(
```
RookieRick: migrated the old namedtuple to a subclass so we could rename `marshal_schema` to `response_body_schema` and add a to-be-deprecated `marshal_schema` property for backward compatibility.  Moved elsewhere in the file to sit with other class defs

flask_rebar/utils.py
```diff
@@ -0,0 +1,124 @@
+"""
```
brockhaywood: is utils the right name for this package? should we be more specific? imo, any utils package ends being a dumping ground when we could just be diligent about structuring things.

RookieRick: Hrmmm - yeah, I wrestled a bit with that internally; given that it struck you as well (and that in the course of this PR it already evolved a bit past what I initially had in mind) it's probably worth a little refactor before we merge

flask_rebar/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+
+    def decorator(f):
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            new = None
+            eol = eol_version
+            if (
+                type(new_func) is tuple
+            ):  # in case somebody inferred this from the way we deprecate params..
```
airstandley: If you are going to do this you probably should also check that `eol_version is None`.

flask_rebar/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    Adapted from https://stackoverflow.com/a/49802489/977046
+    :param aliases: Keyword args in the form {old_param_name = Union[new_param_name, (new_param_name, eol_version)]}
+                    where eol_version is the version in which the alias may case to be recognized
+    :return: function decorator that will apply aliases to param names and raise DeprecationWarning
+    """
+
+    def decorator(f):
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            _rename_kwargs(f.__name__, kwargs, aliases)
```
airstandley: I've never been a big fan of functions updating mutable objects. I'd much prefer if `_rename_kwargs` returned a new `dict` object with the updated mappings.

RookieRick:
> Very nice. There's a few things I have opinions on, but it looks great overall.

Great suggestions; thanks -- I intend to open a minor maint/release PR to address an apparent missing Git tag and release (or pre-release if I can figure out good strategy in short order) so will address these with that one 👍

PR #104, opened 2019-06-12
Deprecation util cleaned up and expanded a bit

RookieRick:
Tightened up some stuff in the deprecation util - was missing a place where tuple-handling should have been adjusted, which pointed out that said tuple-handling really should have been encapsulated in its own helper method, which inspired more thoughts about crazy inputs we might expect and deal with.

Whether this PR represents a badge of honor or a cry for help is left as an exercise for the reader.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.8.0
+current_version = 1.8.1
 commit = True
 tag = True
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,8 @@ Changelog
 =========
 
 
-(unreleased)
-------------
+v1.8.1 (2019-06-14)
+-------------------
 
 Changes
 ~~~~~~~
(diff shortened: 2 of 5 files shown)


RookieRick:
Looks like I introduced a bug (used deepcopy when I should have done a shallow copy).  Will plan to include a bumpversion etc to release 1.8.1 with this one when merging.

twosigmajab:
Sorry I couldn't immediately tell, when would users hit this bug / what's the impact?

RookieRick:
> Sorry I couldn't immediately tell, when would users hit this bug / what's the impact?

The sneaky issue that surfaced here: https://github.com/plangrid/flask-rebar/blob/master/flask_rebar/rebar.py#L82  - The deepcopy in that decorator created a duplicate of the authenticator passed in and substituted it (essentially creating a new `default_authenticator ` on every call).  End result: if you added keys to `default_authenticator` they would be ignored.

twosigmajab:
Thanks for explaining! Fortunately that won’t affect any of the Flask-Rebar users I’m supporting (who are using 1.8.0 as of a couple nights ago), but I’ll upgrade us to 1.8.1 when it’s out.

Issue #94, opened 2019-05-10
Minor refactor: move USE_DEFAULT


RookieRick:
`USE_DEFAULT` is defined in `flask_rebar.authenticators` but its usage has been expanded so it probably belongs in a more generic utility module

retornam:
@RookieRick what are your thoughts on moving this into `utils` so it becomed 
``` from flask_rebar.utils import USE_DEFAULT``` or 
```from flask_rebar.utils.defaults import USE_DEFAULT```

RookieRick:
Yeah I think having a `defaults` module in `utils` (even if it just contains that one dummy class for now) seems like a nice clean way to organize this

PR #107, opened 2019-06-26
chg: Move USE_DEFAULT to utils

retornam:
Fixes https://github.com/plangrid/flask-rebar/issues/94

The USE_DEFAULT class started its beautiful life
in the authenticators file, as the project has grown
its responsiblities have increased. We've decided to
move USE_DEFAULT into a new home under the utils
folder.

We hope USE_DEFAULT enjoys its new home with added
responsibility.


RookieRick:
This lgtm; I'll hold off on merging for a moment (possibly until approx. end of week or early-next as everybody at PlanGrid/ACS is hip-deep in our "Hackweek" event) in case anybody else has particularly strong feelings on this as to defining in `__init__.py` vs in a separate module within the package?  I personally lean toward this package-level definition approach vs defining an additional module just to house a do-nothing class, but I also realize that potentially puts me in direct conflict with certain style guides..  But, on the flip-side, we have at least one other instance (`testing/__init__.py`) in which we already do this.. so.. /shrug?   @airstandley @brockhaywood @joeb1415 @barakalon what do you guys think as longer-term Python vets than yours-truly? 😃

barakalon:
The biggest problem I've run into with defining things in `__init__.py` modules is circular dependencies, but this one feels pretty safe 👍

joeb1415:
I'd prefer separate module imports such as `defaults.py` mentioned in the issue, __init__ is a little hidden sometimes for actual code, but not a blocker.

RookieRick:
Sorry for the delay, @retornam would you mind creating a `defaults.py` in `utils` and refactoring this class there from `__init.py__`?  Probably the cleaner approach in the long-run **_and_** keeps me from introducing more conflict with our internal style guide that strongly encourages empty `__init.py__` practice.  Alternatively I can merge this one and then do a new PR but I figured while this one was still open I'd ask first in case you have available bandwidth 😉

retornam:
@RookieRick  done.

tests/test_rebar.py
```diff
@@ -14,13 +14,13 @@
...
 
-from flask_rebar import HeaderApiKeyAuthenticator, SwaggerV3Generator
-from flask_rebar.authenticators import USE_DEFAULT
 from flask_rebar import messages
+from flask_rebar import HeaderApiKeyAuthenticator, SwaggerV3Generator
 from flask_rebar.compat import set_data_key
 from flask_rebar.rebar import Rebar
 from flask_rebar.rebar import prefix_url
 from flask_rebar.testing import validate_swagger
+from flask_rebar.utils.defaults import USE_DEFAULT
```
joeb1415: Group imports and sort alphabetically

barakalon: `black` should take care of this, right?

Sytten: It doesn't no, there is actually a debate about it

barakalon: Well in that case I don't think we should care. I don't think we should add any additional formatting rules on top of `black`- IMO that would defeat the purpose.

joeb1415: Fair enough. Maybe [isort](https://github.com/timothycrosley/isort) for offering a standard opinion on these?

Sytten: Yeah I looked at isort, it seems nice but I remember seeing some incompatibilities with black (maybe resolved by now).

PR #111, opened 2019-07-12
fix: broke sphinx when we removed requirements.txt

### LINKED ISSUE #110: API docs are broken in Flask-Rebar docs
https://flask-rebar.readthedocs.io/en/latest/api_reference.html currently shows nothing but section headings.

RookieRick:
Resolves #110

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,9 +12,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
+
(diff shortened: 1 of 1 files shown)


airstandley:
@RookieRick I could use some context...

RookieRick:
> @RookieRick I could use some context...

Heh sure - there was a PR a few weeks back that involved getting rid of `requirements.txt` in favor of having dependencies in `setup.py`.  Best I can reason, the loss of "-e ." from that `requirements.txt` caused Sphinx to stop finding `flask_rebar` so this fix (gleaned from https://sphinx-rtd-tutorial.readthedocs.io/en/latest/sphinx-config.html#autodoc-configuration) is to just manually tell sphinx to include the root of the project in sys path.

PR #113, opened 2019-07-24
run exception handlers on sys exit

brockhaywood:


tests/test_errors.py
```diff
@@ -93,6 +97,12 @@ def test_default_500_errors_are_formatted_correctly(self):
         self.assertEqual(resp.content_type, "application/json")
         self.assertEqual(resp.json, {"message": messages.internal_server_error})
 
+    def test_timeouts_log_exceptions(self):
+        # in the wild, gunicorn or nginx will cutoff the wsgi server and return a 502
+        # causing the wsgi server to raise SystemExit
+        with self.assertRaises(SystemExit):
+            self.app.test_client().get("/slow")
+
```
airstandley: Based on the name it seems like we should have a check that confirms the exception was logged?

PR #116, opened 2019-08-20
Avoid breaking changes to marshmallow

RookieRick:
Expected with 3.0 latest release notes.  
Unexpected with 2.20 - minor, but `test_doesnt_break_for_non_object_schema` call to `load` with invalid data now returns `None` instead of `{}` which breaks one of our unit tests


RookieRick:
> agree with @barakalon this seems like a breaking change

Hrmm yeah you guys are right, better to just fix the "broken" test.

PR #117, opened 2019-08-20
Relax overly-sensitive test

RookieRick:
(when updating from 2.19 to 2.20) It appears that marshmallow went from returning an error code and a `data` of `{}` to returning error code and `data` as `None`.  Previously in this test we explicitly checked for `assertEqual(data, {})` Seems reasonable to assume that any sane developer wouldn't be trying to do anything that relies on `data` being an empty dictionary when there's an error code so simply relaxing the test to accept any "falsey" `data` reply.
Will open a separate issue to figure out what we need to do for the 3.0.0 compatibility stuff


Sytten:
The title of the PR is misleading for the changes made, did we forget to actually handle the case in the flask-rebar code? I only see a test...

RookieRick:
> The title of the PR is misleading for the changes made, did we forget to actually handle the case in the flask-rebar code? I only see a test...

I don't believe we ever actually look at/for that particular return value in Rebar; as far as I can tell this test was just a bit overzealous.  Added a bit more to description and renamed this PR to reflect that. 👍

PR #126, opened 2019-09-16
 Accept bare class for schema

### LINKED ISSUE #115: Obscure error message when user forgets to add parentheses.
A simple parentheses omission in schema instantiation like below (see the comment in the `request_body_schema` line) can cause a gnarly swagger generation error message:

Code:
```
from marshmallow import Schema
from flask_rebar import Rebar

class MySchema(Schema):
    ...

rebar = Rebar()
registry = rebar.create_handler_register(...)
@registry.handles(
    rule='/endpoint',
    method='POST',
    request_body_schema=MySchema   # Should be MySchema()
):
    ...
```

Error message:
flask_rebar.swagger_generation.marshmallow_to_swagger.UnregisteredType: No registered type found in method resolution order: (<class 'marshmallow.schema.SchemaMeta'>, <class 'type'>, <class 'object'>)
Registered types: [<class 'marshmallow.fields.Boolean'>, <class 'marshmallow.fields.
Date'>, ...]

I spent ~30 minutes debugging a similar error in my app. It would be great if rebar could suggest to the user to check if they missed any parentheses.

RookieRick:
Fixes #115

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -17,7 +17,6 @@ from copy import copy
 from distutils.version import LooseVersion
 from functools import wraps
 from inspect import isclass
-from types import MethodType
 
 import marshmallow
 from flask import __version__ as flask_version
@@ -471,9 +470,12 @@ class HandlerRegistry(object):
         headers_schema = normalize_schema(headers_schema)
         request_body_schema = normalize_schema(request_body_schema)
         query_string_schema = normalize_schema(query_string_schema)
+
         if response_body_schema:
-            for code, schema in response_body_schema.items():
-                response_body_schema[code] = normalize_schema(schema)
+            response_body_schema = {
+                code: normalize_schema(schema)
+                for (code, schema) in response_body_schema.items()
+            }
 
         # authenticators can be a list of Authenticators, a single Authenticator, USE_DEFAULT, or None
         if isinstance(authenticators, Authenticator) or authenticators is USE_DEFAULT:
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -468,6 +468,7 @@ class HandlerRegistry(object):
             response_body_schema = {200: response_body_schema}
 
         # Fix #115: if we were passed bare classes we'll go ahead and try to instantiate
+        headers_schema = normalize_schema(headers_schema)
         request_body_schema = normalize_schema(request_body_schema)
         query_string_schema = normalize_schema(query_string_schema)
         if response_body_schema:
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -16,7 +16,6 @@ from collections import namedtuple
 from copy import copy
 from distutils.version import LooseVersion
 from functools import wraps
-from inspect import isclass
 
 import marshmallow
 from flask import __version__ as flask_version
@@ -38,6 +37,11 @@ from flask_rebar.utils.deprecation import deprecated, deprecated_parameters
 from flask_rebar.swagger_generation import SwaggerV2Generator
 from flask_rebar.swagger_ui import create_swagger_ui_blueprint
 
+# Deal with maintaining (for now at least) support for 2.7+:
+try:
+    from collections.abc import Mapping  # 3.3+
+except ImportError:
+    from collections import Mapping  # 2.7+
 
 # To catch redirection exceptions, app.errorhandler expects 301 in versions
 # below 0.11.0 but the exception itself in versions greater than 0.11.0.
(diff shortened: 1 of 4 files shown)


flask_rebar/rebar.py
```diff
@@ -16,6 +16,8 @@
 from copy import copy
 from distutils.version import LooseVersion
 from functools import wraps
+from inspect import isclass
+from types import MethodType
```
twosigmajab:
```suggestion
```
unused import

flask_rebar/rebar.py
```diff
@@ -459,9 +462,18 @@ def add_handler(
         :param Type[USE_DEFAULT]|None|str mimetype:
             Content-Type header to add to the response schema
         """
-        if isinstance(response_body_schema, marshmallow.Schema):
+        if isinstance(response_body_schema, marshmallow.Schema) or isclass(
```
twosigmajab:
```suggestion
        if response_body_schema is not None:
```
(Consolidate the further checking and transformation of what was passed in to the new normalize function)

flask_rebar/rebar.py
```diff
@@ -459,9 +462,18 @@ def add_handler(
...
+            response_body_schema
```
twosigmajab:
```suggestion
```

flask_rebar/rebar.py
```diff
@@ -459,9 +462,18 @@ def add_handler(
...
+        ):
```
twosigmajab:
```suggestion
```

flask_rebar/rebar.py
```diff
@@ -459,9 +462,18 @@ def add_handler(
...
             response_body_schema = {200: response_body_schema}
 
+        # Fix #115: if we were passed bare classes we'll go ahead and try to instantiate
+        request_body_schema = normalize_schema(request_body_schema)
+        query_string_schema = normalize_schema(query_string_schema)
+        if response_body_schema:
+            for code, schema in response_body_schema.items():
```
twosigmajab:
```suggestion
            response_body_schema = {code: normalize_schema(schema) for (code, schema) in response_body_schema.items()}
```
clearer as a dict comprehension

flask_rebar/rebar.py
```diff
@@ -459,9 +462,18 @@ def add_handler(
...
+                response_body_schema[code] = normalize_schema(schema)
```
twosigmajab:
```suggestion
```

flask_rebar/utils/request_utils.py
```diff
@@ -92,7 +92,7 @@ def normalize_schema(schema):
     This allows for either an instance of a marshmallow.Schema or the class
     itself to be passed to functions.
     """
-    if not isinstance(schema, marshmallow.Schema):
+    if schema is not None and not isinstance(schema, marshmallow.Schema):
```
twosigmajab:
```suggestion
    if issubclass(schema, marshmallow.Schema):
```

flask_rebar/rebar.py
```diff
@@ -16,6 +16,8 @@
 from copy import copy
 from distutils.version import LooseVersion
 from functools import wraps
+from inspect import isclass
```
twosigmajab:
```suggestion
```
(Would be unused after my suggestion below.)

flask_rebar/rebar.py
RookieRick: Whoops guess we should maybe add Flake8 to the pre-commit hooks :)

flask_rebar/rebar.py
RookieRick: I think we still need this (or a similar) bit of extra checking as the user could have passed either a schema by itself or a dictionary of {response: schema} items; if they passed a schema (or schema class) it still needs to get wrapped in the dictionary.

flask_rebar/utils/request_utils.py
RookieRick: Issue with this is that normalize_schema as it exists will accept:
1. an actual instantiated schema, in which case it just parrots back (those cases would cause an exception with `issubclass` absent some other checking to decide if the thing we have been handed is a class or an object)
2. anything invocable (in which case the current behavior is prob technically incorrect.. as it doesn't enforce that the invocable thing is a class and further a subclass of marshmallow.Schema, but since this has been in the wild for some time, conceivably there could be some evil genius [or lunatic, YMMV] out there passing some kind of generator function etc etc so I'm loathe to break it)

flask_rebar/rebar.py
RookieRick: Per my response on the other convo, we do still need to do the extra check (to detect when we are/aren't passed a dictionary) so leaving this one in for now, but I'm also open to suggestions for a better way to accomplish that check without breaking any backwards compatibility so leaving this one here for now.

flask_rebar/utils/request_utils.py
airstandley: Crap I forgot `issubclass` throws TypeErrors.
It seems like the docstring makes it fairly clear that the method is only intended to work with classes/class instances that derive from Marshmallow.Schema. 

I'd just throw the TypeError it it's not one of those:
```
if not isinstance(schema, marshmallow.Schema) and issubclass(schema, marshamallow.Schema):
```
Or `inspect.isclass` _should_ work if you want to avoid the TypeError.

But it does seems like there are a few equally valid ways you could take this. 🤔

flask_rebar/rebar.py
```diff
@@ -459,9 +461,23 @@ def add_handler(
         :param Type[USE_DEFAULT]|None|str mimetype:
             Content-Type header to add to the response schema
         """
-        if isinstance(response_body_schema, marshmallow.Schema):
+        if isinstance(response_body_schema, marshmallow.Schema) or (
+            isclass(response_body_schema)
+            and issubclass(response_body_schema, marshmallow.Schema)
+        ):
```
airstandley: Any reason not to just use `isinstance(response_body_schema, dict)`?

flask_rebar/rebar.py
twosigmajab: (Oh duh, sorry! Have a followup suggestion, but I'll leave it against the latest version)

flask_rebar/rebar.py
twosigmajab: Was just going to propose something similar! The more generic way to write that would be:
```python
if not isinstance(response_body_schema, collections.abc.Mapping):
    ...
```
The only issue is if the API is meant to support passing in a generator of `(key, value)` pairs as well as some kind of `Mapping`, but probably fair to just document that this case only supports passing in a `dict` and use `if not isinstance(response_body_schema, dict):`

twosigmajab: Also, what do you think of moving this whole check inside the `if response_body_schema:` check below? That would better consolidate all the handling of `response_body_schema` in one place.

Issue #128, opened 2019-09-17
Description for schema fields


artreven:
Hi there. Thanks for the great project!

I do not find a way to provide description for different fields in my request schema. Is it supported?

```
class GetRelationsSchema(RequestSchema):
    """
    Some request description
    """
    text = fields.String(
        required=True,
        description="some field description")
```

results in 
![swagger_field](https://user-images.githubusercontent.com/1529120/65025651-673d2580-d937-11e9-9ab2-0146a5e29e30.png)

I use `SwaggerV3Generator`.

artreven:
I think I found the source of the problem.
The generated JSON description has this part:
```
"parameters": [
          {
            "in": "query",
            "name": "text",
            "required": true,
            "schema": {
              "description": "Seom field description",
              "type": "string"
            }
          }
        ],
```
however, OpenAPI3 spec expects the description to be 1 level higher, see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-6

airstandley:
Technically SwaggerV3Generator follows 3.0.2; not that it matters in this case.

I think this ended up this way because the generated JSON description is not 'necessarily' wrong.

"SchemaObject" can have a description https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object

It's not entirely clear to me whether the intention of passing `description` metadata to Marshmallow.field.Field was to describe the Schema the field represented or to describe the property/parameter.

airstandley:
Based on @barakalon 's endorsement of the ticket I'm going to assume it was intended to be the parameter description 😄

RookieRick:
Fixed by #129

PR #129, opened 2019-09-17
Treat "description" key the same way as "explode" key for query and h…

artreven:
…eader parameters

See issue #128 .

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.10.0
+current_version = 1.10.1
 commit = True
 tag = True
 
--- a/flask_rebar/swagger_generation/swagger_generator_v3.py
+++ b/flask_rebar/swagger_generation/swagger_generator_v3.py
@@ -317,6 +317,7 @@ class SwaggerV3Generator(SwaggerGenerator):
             # Pardon the ugliness.
             # We need the "explode" key to be at the parameters level, not at the schema level.
             explode = jsonschema.pop(sw.explode, None)
+            description = jsonschema.pop(sw.description, None)
 
             parameter = {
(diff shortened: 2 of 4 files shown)


brockhaywood:
@RookieRick Sending this your way. I'm not nearly as familiar with what level of support we have for which versions of openapi.

RookieRick:
> Awesome! Thanks for doing this.
> 
> Is it possible to add a test for this change?

+1 to this, we'll want to add a test case to ensure the generated Swagger matches expectation - @arteven if you can add that, that would be great - if needed though I can prob throw something together for that, just let me know.  And thanks for the contribution!!

airstandley:
It might be worth making a more general method to handle bumping up a list of properties from the `jsonschema` to the `parameter`.

Looking over the 2.0/3.0.2 differences if we ever added support for any of the following fields, then I think they would also want to be "bumped up a level".
v2 & v3: allowEmptyValues
v3 only: deprecated, allowReserved, example, examples

artreven:
Thanks for the feedback. I have modified an existing test "exploded_query_string" to include the description of parameters. Not sure if it is the right way, though, please, have a look.

I am new to gitchangelog. It has output some strange message "Revision '^v1.10.0' is not valid.". Please, also have a look if everything is fine with the changelog.

RookieRick:
> Thanks for the feedback. I have modified an existing test "exploded_query_string" to include the description of parameters. Not sure if it is the right way, though, please, have a look.
> 
> I am new to gitchangelog. It has output some strange message "Revision '^v1.10.0' is not valid.". Please, also have a look if everything is fine with the changelog.

Thanks!  No worries on the version thing - that was a result of a typo'd version number on my part.  I'll take care of that and changelog at same time :)

Issue #125, opened 2019-09-12
Improve Swagger Generation for complicated Authenticators (and OAuth/OpenIDConnect)


airstandley:
Came across another fun Authentication use case that Rebar doesn't currently support.

OpenAPI 2 and 3 both supported AND/OR security requirements. https://swagger.io/docs/specification/authentication/#multiple

The example they give is an API that requires both an API Key and an App ID. 

Writing an Authenticator that does this is simple:

```
from flask import request, g
from flask-rebar import HeaderAPIKeyAuthenticator, errors


class MyAuthenticator(HeaderAPIKeyAuthenticator):
    def __init__(api_key_header, app_id_header, name="myAuth")
        super(MyAuthenticator, self).__init__(api_key_header, name=name)
        self.app_id_header = app_id_header

    def authenticate(self):
        super(MyAuthenticator, self).authenticate()
        if self.app_id_header not in request.headers:
            raise errors.Unauthorized("Missing application id")
        g.authenticated_app_id = request.headers.get(self.app_id_header)


authenticator = MyAuthenticator("X-API-Key", "X-App-ID")
```

The correct swagger spec is also pretty simple

```
"securitySchemes": {
    "apiKey": {
        "type": "apiKey",
        "in": "header",
        "name": "X-API-Key"
    },
    "appID": {
       "type": "apiKey",
        "in": "header",
        "name": "X-App-ID"
    },
}

"security": [
    {"apiKey": [], "appID": []},
]
```

However you're basically stuck writing your own SwaggerGenerator if you want to generate such a spec since our AuthenticatorConverter only supports {name: []} SecurityRequirements and only supports one SecurityDefinition per Authenticator.

Also that {name:[]} SecurityRequirements assumption means that we can't generate valid requirements for OAuth and OpenIDConnect since those should include scope names in the list.

I'm proposing switching to Authenticators registering a AuthenticatorConverter object rather than a single method. An AuthenticatorConverter should have `get_security_schemas` and `get_security_requirements` methods that accept an Authenticator instance, and return a list of Security Scheme Objects and Security Requirement Objects respectively

Sytten:
I don't believe we support AND in our authenticators. But we do support the OR now. What I would really like to add is something like what spring does with the @PreAuthorize (https://docs.spring.io/spring-security/site/docs/3.0.x/reference/el-access.html).
[...]

PR #130, opened 2019-09-17
Improve swagger support for authenticators

airstandley:
BP-778
Issue #125

Makes it so Authenticator converters can define separate methods for getting security requirements and security definitions rather than making assumptions about the requirements based on the definition.

### DIFF
--- a/flask_rebar/swagger_generation/authenticator_to_swagger.py
+++ b/flask_rebar/swagger_generation/authenticator_to_swagger.py
@@ -1,11 +1,7 @@
 from collections import namedtuple
 
 from flask_rebar.utils.deprecation import deprecated, deprecated_parameters
-from flask_rebar.authenticators import (
-    HeaderApiKeyAuthenticator,
-    USE_DEFAULT,
-    Authenticator,
-)
+from flask_rebar.authenticators import HeaderApiKeyAuthenticator, Authenticator
 from .marshmallow_to_swagger import ConverterRegistry
 from . import swagger_words as sw
 
@@ -66,7 +62,7 @@ class HeaderApiKeyConverter(AuthenticatorConverter):
 
     def get_security_requirements(self, obj, context):
         """
-        :param HeaderApiLeyAuthenticator obj:
+        :param HeaderApiKeyAuthenticator obj:
         :param _Context context:
         :return: list
         """
--- a/flask_rebar/swagger_generation/authenticator_to_swagger.py
+++ b/flask_rebar/swagger_generation/authenticator_to_swagger.py
@@ -19,22 +19,76 @@ class AuthenticatorConverter(object):
     """
     Abstract class for objects that convert Authenticator objects to
     security JSONSchema.
+
+    When implementing your own AuthenticatorConverter you will need to:
+
+    1) Set AUTHENTICATOR_TYPE to be your custom Authenticator class.
+
+    2) Configure get_security_scheme to return a map of Security Scheme Objects.
+        You should check for the openapi_version in context.openapi_version;
+        If generation is requested for a version of the OpenAPI specification
+        you do not intend to support,  we recommend raising a NotImplementedError.
+
+    3) Configure get_security_requirements to return a list of Security Requirement Objects.
+        You should check for the openapi_version in context.openapi_version;
+        If generation is requested for a version of the OpenAPI specification
+        you do not intend to support, we recommend raising a NotImplementedError.
+
     """
 
     AUTHENTICATOR_TYPE = None
 
     def get_security_schemes(self, obj, context):
         """
-        Get the security schemes for the provided Authenticator object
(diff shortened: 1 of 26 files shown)


Sytten:
I will do a full review of this (sick this week :( ), I think I might have a few suggestions for you since I built one from scratch.

flask_rebar/swagger_generation/authenticator_to_swagger.py
```diff
@@ -0,0 +1,168 @@
...
+    Abstract class for objects that convert Authenticator objects to
+    security JSONSchema.
+    """
+
+    AUTHENTICATOR_TYPE = None
+
+    def get_security_schemes(self, obj, context):
+        """
+        Get the security schemes for the provided Authenticator object
+        :param flask.authenticators.Authenticator obj:
```
Sytten: Add a space to be consistent

flask_rebar/swagger_generation/authenticator_to_swagger.py
```diff
@@ -0,0 +1,168 @@
...
+        :param _Context context:
+        :rtype: dict
+        """
+        raise NotImplementedError()
+
+    def get_security_requirements(self, obj, context):
```
Sytten: It took me a while to understand what you meant by requirements. I believe we should use `get_security_scopes` to stay consistent with the openapi documentation. Maybe also reference the part of their doc for that.

flask_rebar/swagger_generation/swagger_generator_v2.py
```diff
@@ -116,20 +121,20 @@ def generate(
...
+        for authenticator in authenticators:
+            security_definitions.update(
+                self.authenticator_converter.get_security_schemes(authenticator)
+            )
+
+        default_security = []
+        for authenticator in registry.default_authenticators:
+            default_security.extend(
+                self.authenticator_converter.get_security_requirements(authenticator)
+            )
```
Sytten: So if I am not mistaken we only support the OR case:
```
    security:    # A OR B
      - A
      - B
```
Since OpenAPI supports more than that, I would like to make a case that we also support the combination of auth. Though this is out of scope for this PR.

flask_rebar/swagger_generation/authenticator_to_swagger.py
airstandley: Hmm, that's not good. 
I was trying to be consistent with the OpenApi documentation names, but the purpose of the methods needs to be as clear as possible.

Relevant parts I looked at:

> Definition/Scheme:
>https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-definitions-object
>https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-scheme-object
>https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-scheme-object

>Requirements:
>https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-requirement-object
>https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-requirement-object

@Sytten Do you think adding links  to reference those parts of the docs would be enough to remove confusion?

flask_rebar/swagger_generation/swagger_generator_v2.py
airstandley: My thought was that since Authenticators are classes you can add as much complexity there as you want. (I don't think we ever want to limit the functionality of an authenticator to a validating a single security definition.)

Given that you had Authenticator A and Authenticator B, and for one endpoint you wanted to require A && B, you could already create an Authenticator C which combines A and B.

With this change you can now also define the correct swagger security objects for that endpoint, which is what was not possible before.

So while I agree that it might be nice to have better support for `A AND B`, this should mean that we at least can properly support it; even if it's a bit dirty.

airstandley: (But also I'm kicking myself for not considering that when working on multiple Authenticator support since it probably would not have been too much extra work to support.)

flask_rebar/swagger_generation/authenticator_to_swagger.py
Sytten: Yes, especially since we want users to extend those classes. I was using this doc: https://swagger.io/docs/specification/authentication/#scopes which doesn't mention the word security requirements anywhere :S

flask_rebar/swagger_generation/swagger_generator_v2.py
Sytten: Agreed on that, maybe we should add documentation or examples on how to do that?

airstandley:
Added better doc strings and recipe examples.

@Sytten would love to get your opinion.

Sytten:
Sorry didnt see the message, looks very good to me!

PR #133, opened 2019-09-19
Update authenticators to catch Forbidden exception

Medaragon:
If one authenticator throws a Forbidden Exception, it will not be caught by the current code, and thus, will not check the other authenticators. This is a fix to that problem.


RookieRick:
Adding Andrew as he's been working on authenticator stuff quite a bit recently 👍

Sytten:
Seems legit!

PR #152, opened 2020-01-18
Improve "basics" docs

twosigmajab:
1. Document that 3-tuples can also be returned to include custom headers.
1. Improve "Todo" example based on how such an app would actually be written:
    - Combine `TodoSchema` and `CreateTodoSchema`, so that `TodoSchema` actually includes the description (not just the id), and make `id` a `dump_only` field, so that there's no need for a separate schema for the `create_todo` endpoint's `request_body_schema`.
    - Update `create_todo` to actually use `request_body_schema`.
    - Demonstrate (1.) by including a `Location` header in `create_todo`'s 201 Created response, in compliance with the [HTTP spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201) ("...the [new resource's] location being... the content of the `Location` header").


docs/quickstart/basics.rst
```diff
@@ -159,20 +156,25 @@ The ``response_body_schema`` (previously ``marshal_schema``) argument of ``Handl
 
 In the case of a ``marshmallow.Schema``, that schema is used to ``dump`` the return value of the handler function.
 
-In the case of a dictionary mapping integers to ``marshmallow.Schemas``, the integers are interpreted as status codes, and the handler function must return a tuple of ``(response_body, status_code)``:
+In the case of a dictionary mapping integers to ``marshmallow.Schemas``, the integers are interpreted as status codes, and the handler function must return a tuple like ``(response_body, status_code)``,
+or like ``(response_body, status_code, headers)`` to also include custom headers:
```
airstandley: Nitpick: I could be wrong, but I believe that we still assume a non-tuple response has a 200 status_code. If that is the case we might want to also document that as a valid response here?

docs/quickstart/basics.rst
```diff
@@ -19,7 +19,8 @@ Let's take a look at a very basic API using Flask-Rebar:
 
 
    class TodoSchema(ResponseSchema):
-       id = fields.Integer()
+       id = fields.Integer(dump_only=True)
+       description = fields.String(required=True)
```
Sytten: I personally like having different schema for create and get, but this pattern is also nice so I am not sure what we should recommend to beginners.

docs/quickstart/basics.rst
twosigmajab: Pushed another revision based on this comment. Ended up making what I think are a few other nice improvements to these docs, which I'll call out separately. Thanks for the suggestion!

docs/quickstart/basics.rst
twosigmajab: Check out the latest revision (which actually includes a little more explanation in the docs) and let me know what you think.

When the data for the same type of object is for some reason very different in request bodies compared to response bodies, it makes sense to use different schemas. But in the simpler case (which is maybe more common?) where the only difference is that some fields should only be read-only by users, since they're managed entirely by the database or application, then I think using a single Schema with those fields marked as `dump_only` is a perfect fit. Common examples in my experience are autogenerated `id`s and automatically maintained `updated_at` and `created_at` timestamps, which can all be maintained by a SQL database automatically.

FWIW, this is the pattern I use in the class I've been teaching to beginners and it's been working really well. (For the past year I've been teaching a monthly class at Two Sigma on how to build REST APIs with Flask-Rebar, and by the time they leave the class, students have built a full working CRUD app to manage (you guessed it) their Todo list. 🤣)

twosigmajab:
My pleasure, @airstandley! Note, all Travis jobs passed except one, but that looks like an ephemeral networking failure during the `pip install` (specifically `pip._vendor.urllib3.exceptions.ProtocolError: ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))`). It might be worth prefixing calls that require the network with a good old [`travis_retry`](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) (depending on what happens with #160:).

docs/quickstart/basics.rst
```diff
@@ -60,38 +78,49 @@ Request Body Validation
...
+       # We'll want to return a 201 Created response with a Location header, so use
+       # flask.url_for to get the URL of the new Todo:
+       new_todo_url = flask.url_for(get_todo, id=new_todo.id)
+       response_data = {"id": new_todo.id, "description": new_todo.description}
+       return (response_data, 201, {"Location": new_todo_url})
 
 
-``RequestSchema`` is an extension of ``marshmallow.Schema`` that throws an internal server error if an object is missing a required field. It's usage is optional - a normal Marshmallow schema will also work.
-
 This request schema is passed to ``request_body_schema``, and the handler will now call `marshmallow.Schema.load <http://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.Schema.load>`_ on the request body decoded as JSON. A 400 error with a descriptive error will be returned if validation fails.
```
Sytten: RequestSchema is still important for marshmallow 2 if I am not mistaken, otherwise additional fields are not rejected. What I think we should do in V2 is a single Schema class that includes the RequestSchema and ResponseSchema, but I will open an issue for that.

docs/quickstart/basics.rst
Sytten: Definitely much better with the new comments, I am not sure about the form (explanation before/after code vs inline comments) but that can be improved later.

docs/quickstart/basics.rst
```diff
@@ -8,46 +8,64 @@ Let's take a look at a very basic API using Flask-Rebar:
 
 .. code-block:: python
 
+   # todos.py
+
    from flask import Flask
    from flask_rebar import Rebar
-   from flask_rebar import ResponseSchema
```
twosigmajab: FWIW, I've never used `ResponseSchema` because I've always been the one in full control of my data, and have automated tests that make sure the data I keep in my database stays consistent (validation-wise) with the data my API promises to return. In the interest of keeping this example as basic as possible, while still being as useful as possible, I thought it's better to save Rebar's special `RequestSchema/ResponseSchema` for a later section, in return for demonstrating the other stuff I added here.

docs/quickstart/basics.rst
```diff
@@ -8,46 +8,64 @@ Let's take a look at a very basic API using Flask-Rebar:
...
       rule='/todos/<id>',
       method='GET',
-      response_body_schema=TodoSchema()  # for versions <= 1.7.0, use marshal_schema
+      # Marshal bodies of 200 (default) responses using the TodoSchema.
+      response_body_schema=TodoSchema()
    )
    def get_todo(id):
-       ...
-       return {'id': id}
+       todo = _get_todo_or_404(id)  # Some helper function that queries our database.
```
twosigmajab: Shout out to https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.BaseQuery.first_or_404

docs/quickstart/basics.rst
```diff
@@ -8,46 +8,64 @@ Let's take a look at a very basic API using Flask-Rebar:
...
+       # As with Flask, a 200 response code is assumed when we don't say otherwise.
+       # Return a dictionary with the data for this Todo, which Rebar will marshal
+       # into the response body using our TodoSchema, as we specified above.
+       return {'id': todo.id, 'description': todo.description}
 
    app = Flask(__name__)
    rebar.init_app(app)
 
-   if __name__ == '__main__':
-       app.run()
```
twosigmajab: (`app.run()` has been deprecated/discouraged for [several years](https://stackoverflow.com/a/39088836), and it's just as easy to show best practice here instead, so I fixed that too while I was in here.)

docs/quickstart/basics.rst
```diff
@@ -60,38 +78,49 @@ Request Body Validation
 
 .. code-block:: python
 
-   from flask_rebar import RequestSchema
```
twosigmajab: (FWIW, I haven't yet hit a case where I've wanted Rebar's `RequestSchema`; in all my use cases so far, I've preferred ["being liberal in what I accept and conservative in what I send"](https://en.wikipedia.org/wiki/Robustness_principle), at least as far as unrecognized fields are concerned.)

docs/quickstart/basics.rst
```diff
@@ -60,38 +78,49 @@ Request Body Validation
...
-
-
-   class CreateTodoSchema(RequestSchema):
-       description = fields.String(required=True)
```
twosigmajab: (Now reusing `TodoSchema` for both requests and responses)

docs/quickstart/basics.rst
twosigmajab: See https://github.com/plangrid/flask-rebar/pull/152#discussion_r373322964

docs/quickstart/basics.rst
Sytten: I tend to agree on that one but I think we should move toward #161

docs/quickstart/basics.rst
Sytten: I put it everywhere since I prefer a break fast strategy with a REST API. For example, if for some reason you have a breaking change in the API, you don't want to have clients sending requests thinking that all went well and it applied some changes but in reality it didn't (and no alert since its a 200).

twosigmajab: (To be clear, I'm by no means saying it's always wrong to fail hard for unrecognized fields, just that it's definitely not always right. Maybe more citation needed, so just did a quick search and immediately turned up something interesting: https://github.com/zalando/restful-api-guidelines/pull/248/files#diff-bf72c91a3bee08729395c486d36a5cc2R54-R58 – this topic is deep!)

Sytten: I really like this website, thanks for sharing. I think we should try to encourage some of it's principles while allowing divergences because they are quite strict and hard to meet for a lot of people.

twosigmajab: So looking with fresh eyes today... That API Guidelines link I posted last night is saying that the Robustness Principle applies on the client side, but on the server side, it doesn't!:
> Unknown input fields in payload or URL should not be ignored; servers should provide error feedback to clients via an HTTP 400 response code.

This is not what I thought was best practice, so I'll definitely be doing more research on this! If this API Guidelines doc is correct, then it's quite a validation of `flask_rebar`'s `RequestSchema`, and I'll correct my understanding, my code, and my teaching accordingly!

In the meantime, the current revision of this PR currently is still removing the use of `RequestSchema` for the example `request_body_schema`s. That means going from conforming to these guidelines to not conforming to them. I really don't want to be responsible for, in the process of trying to improve the docs' demonstration of best practices, actually making them _worse_ in this particular way.

At the same time, I'm very keen to see #161 happen, because I do think that having two different schemas (request and response) for the same type of object is not always necessary, and when it isn't, as in this Todo example, it adds unnecessary complexity to our docs. Shall we wait for #161 to be resolved so we can incorporate the outcome into these changes before we merge them?

Thank you @sytten for prompting this!

Sytten: We can wait, I will try to prioritize this weekend of monday. Though I am not sure if we want to only do that in V2 or backport it to V1.

airstandley:
@twosigmajab I really like most the changes!

I'm against removing all references to `RequestSchema/ResponseSchema`; I'm completely pro adding documentation that discusses different philosophies on API design and when `Schema` may be more appropriate than `RequestSchema``ResponseSchema`, but for our basic 'QuickStart' guide I am strongly in favor of suggesting `RequestSchema/ResponseSchema`.

docs/quickstart/basics.rst
```diff
@@ -8,46 +8,64 @@ Let's take a look at a very basic API using Flask-Rebar:
...
-       app.run()
+
+Now run your Flask app `as usual <https://flask.palletsprojects.com/en/1.1.x/quickstart/>`__,
+e.g.:
+
+.. code-block:: bash
+
+   export FLASK_APP=todos.py
+   flask run
+
```
airstandley: I'm wondering if we should split this into two sections, "Creating a Flask App with Rebar" and "Registering a Handler"? As is this portion on running the app feels out of place and I think makes the lines that follow a bit confusing.

jab:
I think there are still some good improvements here that would be beneficial to merge, but I no longer work at Two Sigma, nor with Flask-Rebar, and at this point I would have to hand this off to the maintainers (or another external contributor). Hopefully someone is still interested in making these improvements.

PR #154, opened 2020-01-25
FIX: Use nullable to specify that the v3 value may be null

mbierma:
According to the [OpenAPI v3 specifications](https://swagger.io/docs/specification/data-models/data-types/#null):

> OpenAPI 3.0 does not have an explicit null type as in JSON Schema, but you can use nullable: true to specify that the value may be null. Note that null is different from an empty string "".

Currently, when `allow_none=True` is used to generate a v3 specification, the schema generated will look something like:

`{"anyOf": [{"type": "null"}, {"type": "integer"}]}`

Instead, it should look like:

`{"type": "integer", "nullable": True}`

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -277,7 +277,7 @@ class FieldConverter(MarshmallowConverter):
                     )
 
         if context.openapi_version == 3 and obj.allow_none:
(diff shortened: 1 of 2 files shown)


Sytten:
Good catch, this seems good to me!

airstandley:
Thanks for the fix mbierma!

Coincidentally I was just talking with another dev about using OpenApi v3 for a project given the built-in support for nulls. This would have been a nasty surprise.

Issue #149, opened 2020-01-08
Philosophical (?) question on support for defunct Python versions..


RookieRick:
So in the course of releasing 1.12 this evening I ran into a little snag..  PyYAML (which we were getting as a transient requirement of another library) after version 5.2 dropped support for Python 3.4 (which reached EOL in March 2019).  This in turn caused our Travis build to fail because our Travis config still includes Python 3.4.  My hamfisted workaround was to pin PyYAML <= 5.2 for now..

So the philosophical question is: Would you as a Python developer consider it a "breaking change" if instead of pinning PyYAML to maintain support for 3.4, I had simply removed Travis build vs. 3.4?  

My gut (and heart) say we shouldn't need to do a major version bump every time a 5+ year-old version of Python reaches EOL, so pinning PyYAML feels kludgey, and I would personally lean toward just no longer including 3.4 as "formally supported" in Flask-Rebar (since we don't control when Python versions reach EOL), but as I'm still relatively new to managing a significant OSS project, I wanted to solicit some opinions. :)  Seems like other libraries/apps (like PyYAML for example lol) take the approach of dropping old Python versions in minor version releases.

@barakalon @airstandley @sytten and/or anybody else have thoughts/experience from other projects on this?

P.S. @barakalon I haven't forgotten your suggestion to start doing beta/release candidates, nor does it escape me that this would have been a prime example of a great use case for that.  I swear, I'm going to work that into the process.... 😂

RookieRick:
I should add that I have fewer qualms with tying the dumping of major versions of Python (i.e., our hopefully-soon-to-be release 2.0 that will only support Py3) to a major version bump

Sytten:
As long as we don't change the API, a major is not necessary to dump support for a python version. See [Django for example](https://docs.djangoproject.com/en/3.0/faq/install/), they stopped supporting Python 3.4 at version 2.1. We should actually start to prepare to drop 3.5 (for which I am very excited because it will allow us to use async stuff).

RookieRick:
Nice - I like what they did with their explicit table there too.

barakalon:
Philosophical (?) response:

Let's not forget that we're using semantic versioning. Here's a description from https://semver.org/:

>In the world of software management there exists a dreaded place called “dependency hell.” The bigger your system grows and the more packages you integrate into your software, the more likely you are to find yourself, one day, in this pit of despair.
>
[...]

PR #160, opened 2020-01-30
add workflows for github actions

gtmanfred:
I removed python3.4 because it is end of lifed and doesn't exist on
github.

I also removed marshmallow 3.0.0rc5 since newer 3 versions have changed
the api.  Also, marshmallow 3.0 seems to have dropped support for
python2.


.travis.yml
```diff
@@ -44,6 +44,7 @@ jobs:
       install: skip
 
       deploy:
+        distributions: "sdist bdist_wheel"
```
gtmanfred: At the very least, we should definitely add `bdist_wheel` and publish a universal wheel, instead of only the sdist.

gtmanfred:
Also, github actions run 20 instances in parallel, so the test suite for the whole matrix runs in about 12 minutes, instead of the 12 minutes per test and 40 minutes total on travis.

gtmanfred:
Here is the PR where I was testing these actions on my own branch.

https://github.com/gtmanfred/flask-rebar/pull/1

gtmanfred:
If we move to github actions instead of travis, we will need to update the secret for the password to pypi for releases.

.travis.yml
Sytten: do we really need to though? we don't have compiled code, so it would pretty much be a sdist.

.github/workflows/pullrequests.yml
```diff
@@ -0,0 +1,59 @@
...
+        python:
+        - 2.7
+        - 3.5
+        - 3.6
+        - 3.7
+        - 3.8
+        libraries:
+        - marshmallow=='2.13.*' flask=='0.10.*'
+        - marshmallow=='2.14.*' flask=='0.11.*'
+        - marshmallow=='2.15.*' flask=='0.12.*'
```
Sytten: I think we should drop support for flask<1.0 in v2 @RookieRick @twosigmajab

.github/workflows/pullrequests.yml
```diff
@@ -0,0 +1,59 @@
...
+    - uses: actions/checkout@v1
+    - name: Set up Python:${{ matrix.python }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python }}
+    - name: "Test with ${{matrix.libraries}}"
+      run: |
+        pip install -U pip
+        pip install '.[dev]' ${{matrix.libraries}}
+        pip freeze
```
Sytten: A caching step could be used here to speed up the process

.github/workflows/pullrequests.yml
```diff
@@ -0,0 +1,59 @@
...
+      with:
+        python-version: 3.7
+    - name: "Install black"
+      run: |
+        python -m pip install -U pip
+        python -m pip install black
+        python -m pip freeze
+    - name: Run black
+      run: |
+        python -m black --check --diff .
```
Sytten: I would like to add isort too, but that's out of scope for this PR @RookieRick

.github/workflows/tag.yml
```diff
@@ -0,0 +1,28 @@
...
+    types:
+    - created
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.7
```
Sytten: Why not 3.8?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,28 @@
...
+      with:
+        python-version: 3.7
+    - name: Install pep517
+      run: |
+        python -m pip install pep517 --user
+    - name: Build a binary wheel and a source tarball
+      run: |
+        python -m pep517.build .
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.event.ref, 'refs/tags')
```
Sytten: I am curious to know why this if is necessary

.github/workflows/tag.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install pep517
+      run: |
+        python -m pip install pep517 --user
+    - name: Build a binary wheel and a source tarball
+      run: |
+        python -m pep517.build .
```
Sytten: I don't know about PEP517, can you explain a bit the choice vs others like poetry, flit and basic setup.py? Since the project isn't using new format from PEP517/518 I don't really see the benefit.

.github/workflows/tag.yml
gtmanfred: It was just the one that was used in travis, and it doesn't matter, since the wheel will be universal and pythoon agnostic.

.github/workflows/tag.yml
gtmanfred: I am not sure what it is used for right this second, it is just what is recommended by pypa https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

.github/workflows/tag.yml
gtmanfred: This still uses setup.py.

Python 517 just specifies the build tools section of the pyproject.toml, which tells everyone how the package should be built into an sdist and/or a wheel.  So if we ever move to flit/poetry this build method will still work.

.github/workflows/tag.yml
Sytten: Ho I see, in the example they do it on `push` so they only release to Pypi if the commit is tag. We don't need this since we do it on release created :)

gtmanfred: Ahh yes, so we shouldn't need it.  I do remember that from earlier today, but i got distracted and forgot to remove this line.

.github/workflows/pullrequests.yml
RookieRick: Seems reasonable - I'm tempted to say we do a "soft drop".. i.e., we could leave `"Flask>=0.10,<2"` in setup.py `install_requires` and just remove the explicit test steps here against pre 1.0.  I don't know if there are any (sane) people in the world using pre-1.0 versions of Flask **_and_** using Flask-Rebar, but if they exist, I don't want to break their stuff unless we really have to (i.e., if we made some change that relied on a Flask 1.0+ feature making us no longer compatible with Flask 0.* we could update setup.py as well - and per our discussion my #149 we could justify  doing so with just a minor version bump ;) )

.github/workflows/pullrequests.yml
```diff
@@ -0,0 +1,59 @@
+name: flask-rebar Pull Request Tests
+
+on:
+- pull_request
+
+jobs:
+  tests:
+    name: Testing on Python ${{ matrix.python-version }} and ${{ matrix.os }}
```
RookieRick: Looks like this should be `{{ matrix.python }}` (and `matrix.os` is always blank so we could just leave that off)?

RookieRick:
This looks pretty sweet - couple of lingering questions around black's "perpetual beta" status and the inner-workings logistics for PyPi publishing.  Will hit up you up offline @gtmanfred to chat about those. Thanks!!!

.travis.yml
gtmanfred: the nice part about a wheel is that it never needs to run setup.py, it just lays down the file, so it is faster to install as well.

gtmanfred:
well thats no good.  I broke the pr.

PR #168, opened 2020-02-08
Pin werkzeug

### LINKED ISSUE #166: pin werkzeug <1.0
Werkzeug, a sub-dependency of flask, released a major version 1.0.0 today:
https://pypi.org/project/Werkzeug/#history

This is breaking a few things on some of our builds. Looks like Flask does not pin a major version of werkzeug:
https://github.com/pallets/flask/blob/master/setup.py#L56

There is some discussion in the flask discord about it but no issues or discussion of a resolution yet
https://discordapp.com/channels/531221516914917387/531221866145251366

Can we pin `werkzeug^=0.16` in rebar for now?

Sytten:
Fixes #166 until we change Flask-Testing or it gets an update.
Maybe we could put this pin only in dev dependencies?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,11 @@ if __name__ == "__main__":
         packages=find_packages(exclude=("test*", "examples")),
         include_package_data=True,
         extras_require={"dev": development},
-        install_requires=["Flask>=0.10,<2", "marshmallow>=2.13,<3"],
+        install_requires=[
+            "Flask>=0.10,<2",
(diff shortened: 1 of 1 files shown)


gtmanfred:
Instead of pinning werkzeug, I removed the dependency on Flask-Testing

https://github.com/plangrid/flask-rebar/pull/172

airstandley:
Dependency on Flask-Testing is removed. Closing this out!

Issue #137, opened 2019-10-05
ResponseSchema is missing constructor parameters


Sytten:
I often use the `many=True` for `Schema` in the returned body, but it's not possible with the `ResponseSchema` class because it doesn't implement a constructor.

airstandley:
Hey Sytten,
  `ResponseSchema` does not need to implement it's own` __init__`, it will inherit `Schema`'s

You should be able to register handlers with many=True schemas:
```
@registry.handles(
   response_body_schema=ResponseSchema(many=True),
)
def my_handler():
     pass
```

Hope that helps!

Sytten:
Hum yeah it must work because one of the parent doesn't have a constructor. Just pycharm complaining about it.

RookieRick:
Hey, just doing some cleanup after being drawn away from Rebar for far too long ;)  Looks like this one is resolved; if not feel free to reopen

Sytten:
Sure, its just annoying in pycharm but it works

airstandley:
I will admit, that sounds pretty obnoxious.

PR #169, opened 2020-02-08
Feature/strict schema

### LINKED ISSUE #161: Refactor RequestSchema and ResponseSchema into a single Schema
- `RequestSchema` is only needed in v2 to reject additional fields
- `ResponseSchema` is currently not playing too well with IDE (pycharm *cough*) because its missing a constructor with the kwargs (see https://github.com/plangrid/flask-rebar/issues/137)
- Having two to inherit from 2 schemas is not that nice since we use them a lot in requests and responses (see PR #152)
- It's easy to forget and just use Schema from marshmallow (our doc does a poor job of recommending them too)

For those reasons I think we should refactor that into a single class that we use in all the tutorials and doc.

Sytten:
Fixes #161 
This adds a new `StrictSchema` which we should recommend for new users in #152. It's a simple combination of `RequestSchema` and `ResponseSchema`. This is also interesting if we want to extend some functionality of marshmallow down the road.

I also added a simple constructor to `ResponseSchema` because of #137. A future improvement (if/when we drop marshmallow v2) would be to have all the keyword arguments so the IDE can pick it up and suggest them to the user. At least now it doesn't show up as an error.

### DIFF
--- a/flask_rebar/validation.py
+++ b/flask_rebar/validation.py
@@ -121,6 +121,10 @@ class ResponseSchema(ActuallyRequireOnDumpMixin, Schema):
         super().__init__(**kwargs)
 
 
+class StrictSchema(RequestSchema, ResponseSchema):
+    pass
+
+
 class Error(Schema):
     message = fields.String(required=True)
     errors = fields.Dict(required=False)
--- a/flask_rebar/validation.py
+++ b/flask_rebar/validation.py
@@ -115,8 +115,10 @@ else:
     RequestSchema = Schema
 
 
+# Define a constructor to please some IDE type checking
 class ResponseSchema(ActuallyRequireOnDumpMixin, Schema):
-    pass
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
 
 
 class Error(Schema):
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -20,7 +20,7 @@ from flask_rebar import compat
 from flask_rebar import messages
 from flask_rebar.utils.request_utils import normalize_schema
 from tests.helpers import skip_if_marshmallow_not_v2
-from flask_rebar.validation import ActuallyRequireOnDumpMixin
+from flask_rebar.validation import ActuallyRequireOnDumpMixin, StrictSchema
 from flask_rebar.validation import CommaSeparatedList
 from flask_rebar.validation import DisallowExtraFieldsMixin
 from flask_rebar.validation import QueryParamList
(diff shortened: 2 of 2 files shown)


flask_rebar/validation.py
```diff
@@ -115,7 +115,13 @@ def disallow_extra_fields(self, processed_data, original_data):
     RequestSchema = Schema
 
 
+# Define a constructor to please some IDE type checking
 class ResponseSchema(ActuallyRequireOnDumpMixin, Schema):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
jab: Function call overhead in python is significant. It’s best practice not to introduce this extra no-op function call that reduces performance just for some IDEs that aren’t currently finding the inherited method. Can you follow up with your IDE vendor instead of adding this here?

flask_rebar/validation.py
```diff
@@ -115,7 +115,13 @@ def disallow_extra_fields(self, processed_data, original_data):
...
+
+
+class StrictSchema(RequestSchema, ResponseSchema):
     pass
```
jab: Missing docstring for public class.

Once added, the `pass` will be unnecessary.

tests/test_validation.py
```diff
@@ -117,6 +117,30 @@ def test_required_failed_validate(self):
         self.assertIn("one_of_validation", ctx.exception.messages)
 
 
+class ActuallyStrictSchema(StrictSchema):
```
jab: `TestStrictSchema` or `MyStrictSchema`? Shout out to #170.

flask_rebar/validation.py
```diff
@@ -115,7 +115,13 @@ def disallow_extra_fields(self, processed_data, original_data):
     RequestSchema = Schema
 
 
+# Define a constructor to please some IDE type checking
 class ResponseSchema(ActuallyRequireOnDumpMixin, Schema):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+
+class StrictSchema(RequestSchema, ResponseSchema):
```
jab: - Better to inherit the mixin classes directly
- I wonder if there’s a better name than StrictSchema for this, since strict is overloaded with other meaning in Marshmallow v2 (complicated by how Rebar currently already always does a strict load with unstrict Marshmallow v2 schemas, i.e. raising rather than returning a value with errors in it)

jab:
- should we deprecate Request and ResponseSchema at the same time?
- docs and changelog?

Sytten:
- no we have some use cases for them as an advanced feature
- true my bad

airstandley:
@Sytten Thanks for working on this. It looks great, but I would echo the comments about wanting the no-op __init__ removed and documentation updates being needed.

gtmanfred:
Looks like we have some merge conflicts with this pr. I am working on cleaning up the backlog for flask rebar, so if i can help, please let me know.

Issue #91, opened 2019-05-08
Let's talk about content-negotiation


RookieRick:
Following up on recent PRs, figured I would open an Issue for now where we can kick some ideas around on the topic of content negotiation.

A few quick thoughts to kick things off.  I would characterize our options as:
1. Do nothing (ok now that's out of the way)
2. Build our own content negotiation
3. Add a dependency on some existing content negotiation library (I haven't looked deeply into these yet but found a couple like Flask-Accept, Flask-Negotiate, Flask-Negotiation
4. Punt content-type negotiation and add some recipes/examples that demonstrate how one could integrate Flask-Rebar with things like those in item 3.

Pros/cons I see of these:
1. Duh, no good.
2. Pro: We can make it work exactly how we want.  Con:  Risk of reinventing the wheel
3. Pro:  Citizenship badge for OSS :).  Con:  I like the fact that we are only dependent on the two core technologies that we are combining in Flask and Marshmallow and am hesitant to add to the Python equivalent of DLL-Hell.
4. Pro:  Still good OSS citizens (and without choosing sides - allow people to stick with what they like if they already have a preference).  Con:  Pushes some of the DLL-hell burden to our users

If I was forced to rank these I'd probably say 2 & 4 are a virtual tie for me, followed by 3, and then 1.  Definitely interested to hear thoughts from our other contributors and users.

airstandley:
@RookieRick I think there may be two separate discussions. 
1) Should endpoints registered with rebar be allowed to return anything other than a json encoded body in the response?
2) If rebar endpoints can return multiple types of media in the response body, then how do we handle content headers?

We might want to agree on the first point before discussing the second?

RookieRick:
> 1. Should endpoints registered with rebar be allowed to return anything other than a json encoded body in the response?

My $0.02 on this is that we should not _impose_ json-only unless we absolutely must, in the interest of keeping rebar a generic broadly-usable extension of Flask.  If we restrict to json with no way to override that, we risk blocking otherwise-valid uses of Flask.  I'd advocate for treating json as our default (and possibly only "native" supported i.e. my option 4) while still ideally allowing flexibility for somebody that wants to return HTML or CSV or whatever suits their use case.

barakalon:
> 1. Should endpoints registered with rebar be allowed to return anything other than a json encoded body in the response?

[...]

PR #171, opened 2020-02-09
WIP: Content type specification

mbierma:
This is a quick and dirty proof of concept to allow for the specification of content type other than `application/json` for OpenAPI 3 generation. Currently, it is possible to create an endpoint without specifying the `response_body_schema`, and then returning a `Response` object with the desired content type:

```
@registry.handles(
    rule='/todos',
    method='GET',
    query_string_schema=GetTodosQueryStringSchema(),
)
def get_todos():
    return send_file(pdf, as_attachment=True, mimetype='application/pdf')
```

However, in existing implementation, the generated swagger specification does not reflect the correct content type. This PR is intended to demonstrate a way to allow for the specification of content type using a `__content_type__` magic attribute. Additionally, a primitive `response_body_schema` type is used to allow for responses that are not json objects.

```
class PDF(m.fields.String):
    __content_type__ = "application/pdf"

@registry.handles(
    rule='/todos',
    method='GET',
    query_string_schema=GetTodosQueryStringSchema(),
    response_body_schema=PDF
)
def get_todos():
    return send_file(pdf, as_attachment=True, mimetype='application/pdf')
```

> Possibly related to #91

### DIFF
--- a/flask_rebar/swagger_generation/generator_utils.py
+++ b/flask_rebar/swagger_generation/generator_utils.py
@@ -53,7 +53,11 @@ def get_ref_schema(base, schema):
     :return:
     """
     ref = {sw.ref: create_ref(base, get_swagger_title(schema))}
-    return ref if not schema.many else {sw.type_: sw.array, sw.items: ref}
+    return (
+        ref
+        if not getattr(schema, "many", None)
+        else {sw.type_: sw.array, sw.items: ref}
+    )
 
 
 def get_response_description(schema):
--- a/flask_rebar/swagger_generation/swagger_generator_v3.py
+++ b/flask_rebar/swagger_generation/swagger_generator_v3.py
@@ -182,14 +182,7 @@ class SwaggerV3Generator(SwaggerGenerator):
                 if d.response_body_schema:
                     for status_code, schema in d.response_body_schema.items():
                         if schema is not None:
-                            if not getattr(schema, "__content_type__", None):
-                                response_definition = self._get_response_definition(
-                                    schema
-                                )
-                            else:
-                                response_definition = self._get_response_definition(
-                                    schema, schema.__content_type__
-                                )
+                            response_definition = self._get_response_definition(schema)
                         else:
                             response_definition = {sw.description: "No response body."}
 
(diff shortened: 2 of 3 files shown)


mbierma:
Mostly just looking for preliminary feedback. The main intent of this PR is to:

- Generate the correct OpenAPI 3 specification when the content type is not `application/json`
    - Account for the case where a single endpoint may return different content types based on the return code. For example, there may be consistent `application/json` error responses across all endpoints, even for ones that return a csv/file/etc.

Since this deals with response content types, and #91 is dealing with content negotiation, it would be great if there was a consistent interface between the code here and what you're planning for the future

badam5:
@RookieRick @airstandley Tagging you folks here, since you may have more context than I do regarding content types! Any thoughts?

airstandley:
@mbierma The example you've given uses a Field rather than a Schema which presents all kinds of issues. I'm assuming that's a small mistake and you intended for PDF to be a marshmallow.Schema or some class that implements the `dump` interface. I'd be very interested if you could provide an example that actually handled validation (some assertion that pdf data is being returned). 

The approach you are taking has some merit, but I think it would be helpful if you included a working example of a simple handler which returned a valid pdf when hooked up to a test app.

mbierma:
@airstandley thanks for the feedback. I'm going to be working on a better example

I am definitely in agreement that it would be much more useful to introduce a way to support multiple content-types. Do you have any recommendations on how to specify something like that? The first things that come to mind are

```
@registry.handles(rule="/foo/media", method="GET",
                  response_body_schema={200: [MP4(), OGG(), Quicktime()]})
```

or 

```
@registry.handles(rule="/foo/media", method="GET",
                  response_body_schema={200: {'video/mp4': MP4(),
                                              'video/ogg': OGG(),
                                              'video/quicktime': Quicktime()}})
```

For the first case, there'd likely be some validation that each response schema in the list contains a unique `__content_type__` magic attribute. For the second example, the magic attribute could be done away with, since the content-type is specified in the keys.

airstandley:
I haven't checked the 2.0 spec, but 3.0 specs `content` as a map so I'd say that 
```
@registry.handles(rule="/foo/media", method="GET",
                  response_body_schema={200: {'video/mp4': MP4(),
                                              'video/ogg': OGG(),
                                              'video/quicktime': Quicktime()}})
```
seems like a great approach and then we can fallback on the old behaviour and assume 'application/json' if just a schema is given.

gtmanfred:
Those defaults sound good to me. Is there any chance that you still plan on doing this work @mbierma ?

Thanks,
Daniel

PR #172, opened 2020-02-10
drop Flask-Testing dependency for testing

### LINKED ISSUE #166: pin werkzeug <1.0

gtmanfred:
Flask-Testing is broken with Werkzeug 1.0.0.  It also appears that a
patch was made for Flask-Testing 4 months ago, and hasn't been merged,
so it should probably be considered unmaintained.
    
Since we are using pytest anyway, we can just provide our own app
fixture for testing and just use that.

Fixes #166


airstandley:
@gtmanfred Do we need to add the pytest-flask dependency? I'm fairly certain we should be able to remove the requirement without replacing it with another dependency since we aren't really relying on any of the features beyond a fixture app creation.

gtmanfred:
I would be down to remove the dependency all together.  It is a small enough change to just make the fixture ourselves and use the test client.

gtmanfred:
Yup, looks like it should already work, just straight up dropping that dependency.

airstandley:
Sweet. In that case can we split this into two separate issues, removing the dependency and "pytestifing" the tests?

(I know I'm in a tiny minority here, but I really don't like pytest's RTI fixtures. Gives me flashbacks to  nightmares with DUnitX.)

gtmanfred:
I would rather not split that. The only tests that I pytestified are the ones that use fixtures, because I wanted to avoid mixing the fixture + self arguments.

airstandley:
Split out the changes needed to drop Flask-Testing #173

airstandley:
@gtmanfred FWIW I think you great work here "pytestifying" our tests if we want to move in that direction.

gtmanfred:
:+1: no worries I just didn't want to split it out.

Also, you are probably going to need to grab the changes to .travis.yml to get the tests to pass.

And the JSONResponse changes because flask < 1.0.0 does not have .json on the response object.

PR #173, opened 2020-02-11
Remove Flask-Testing dependency.

airstandley:


airstandley:
Yup, we don't actually use any of the features

gtmanfred:
You should be able to just make a base testcase, and use that for all of these, but once the self.app is created, just update the response_class.

https://github.com/plangrid/flask-rebar/pull/172/commits/9ade6d52a686c3caab89a7e54205da0aa30985e6

PR #177, opened 2020-02-20
Add github actions to workflow

gtmanfred:
I removed python3.4 because it is end of lifed and doesn't exist on
github.

Continuation of #160


gtmanfred:
Here is the github actions running on my repo https://github.com/gtmanfred/flask-rebar/runs/458269940?check_suite_focus=true

gtmanfred:
This will cause travis to not run, but once it is merged future prs that are rebased will run github actions.

PR #178, opened 2020-02-20
Load modules where handlers are located using the registry

gtmanfred:
This prevents all of the F401 unused-import messages from flake8, by
letting rebar programatically load the handlers, instead of importing
unused modules.

I don't have a good way to test this right this second, but I am thinking about it.

I think the easiest way to do it would be to redo the example todo app, but 1) don't want to overwrite that git history, and 2) the example app todo.py file reads really well as a "how to get started with flask-rebar" in one file.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -216,6 +216,38 @@ The preceeding example makes `all` fields from ``CreateTodoSchema`` optional in 
 			super(UpdateTodoSchema, self).__init__(partial=partial_arg, **super_kwargs)
 
 
+Structuring Larger Projects
+===========================
 
+When structuring larger projects, there may be more than one directory which has modules which will be loaded into the ``Registry``.  They still need to be imported, but if there is nothing else to do with them, they aren't used. Instead Flask-Rebar can import those modules.
 
+Let's take a look at this setup:
 
+.. code-block::
+
+	application/
+	├── app.py
+	├── config.py
+	├── database.py
+	├── rebar.py
+	├── handlers
+	│   └── create.py
+	├── models
+	│   └── data.py
+	└── schemas
+		├── create.py
+		└── sync.py
+
+Inside of rebar.py we can find this:
+
+.. code-block:: python
+
+	from flask_rebar import Rebar
+
+	rebar = Rebar()
+	registry = rebar.create_handler_registry(handlers='application.handlers')
+	registry.load_handlers(recursive=False)
+
+The ``handlers`` argument to the :class:`flask_rebar.rebar.HandlerRegistry` will be used to look in the ``application.handlers`` package and load all modules that it finds in there, but not recursively.  This will result in all of the handlers getting imported, so that they can be added to the registry, without a having to add a new import to ``application.app`` each time a new file is added to ``application/handlers/``
+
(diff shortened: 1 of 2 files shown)


docs/recipes.rst
```diff
@@ -216,6 +216,38 @@ The preceeding example makes `all` fields from ``CreateTodoSchema`` optional in
...
+		└── sync.py
+
+Inside of rebar.py we can find this:
+
+.. code-block:: python
+
+	from flask_rebar import Rebar
 
+	rebar = Rebar()
+	registry = rebar.create_handler_registry(handlers='application.handlers')
```
barakalon: Can `handlers` just be passed in at the same time that `load_handlers` is called?

gtmanfred: Not in the current implementation, it is currently an attribute that lives in the registry, but there isn't a reason that it couldn't be changed.

My plan was to eventually use the handlers attribute for finding blueprint like objects, so my current preference is to implement it like this so that I don't have to make a backwards incompatible change in the future.

flask_rebar/rebar.py
```diff
@@ -333,6 +338,7 @@ def __init__(
         self.swagger_generator = swagger_generator or SwaggerV2Generator()
         self.swagger_path = swagger_path
         self.swagger_ui_path = swagger_ui_path
+        self.handlers = handlers if isinstance(handlers, list) else [handlers]
```
badam5: Wouldn't this lead to `[None]` in the default case rather than `[]`?

flask_rebar/rebar.py
```diff
@@ -357,6 +363,17 @@ def set_default_authenticators(self, authenticators):
...
+    def load_handlers(self, recursive=False):
+        """
+        Load modules in ``self.handlers`` package
+        :param boolean recursive:
+            recursively look through packages.
+            Default: False
+        """
+        for package in self.handlers:
+            for handler in find_modules(package, recursive=recursive):
+                importlib.import_module(handler)
```
badam5: Should there be any error handling here or are we ok with just bailing out if we don't get the right input?

docs/recipes.rst
badam5: Not sure what Barak is thinking exactly, but in this example, I'd also like the handler module to be passed directly to `load_handlers`. Though I see why you want the attribute on the registry itself.

Potentially related: is there a reason to have an explicit function to call for this rather than having the modules imported implicitly at some point during `init_app`?

gtmanfred: init_app would work for me as well.  But i think I would still like it to be an attribute on the registry, so that there is a saved record of what is going to be loaded at init time.

badam5: Yeah, makes sense to me. I brought up implicitly loading the modules instead, because that might be a way to keep the handlers attribute on the registry while eliminating the extra explicit call to `load_handlers` which might not read as well without referencing the handlers. That's what I'm thinking anyway.

gtmanfred: :+1: yeah sounds good.

gtmanfred:
I am going to revamp this PR at some point when I get time, but i will probably just start over since it isn't that much code. Thanks for the feedback yall.

PR #182, opened 2020-03-13
Fixes for oneof on _flatten

### LINKED ISSUE #181: Definition lost on swagger_generator for one_of schemas.
The bug consists in the fact that in the `_flatten` function, the updated schema is only updated on the definitions when the type=object.  Here: https://github.com/plangrid/flask-rebar/blob/master/flask_rebar/swagger_generation/generator_utils.py#L134

That `if sw.title in schema:` block should be applied for all the cases (including `elif subschema_keyword:`, and for arrays as well).

This bug also happens on flask-rebar v2, but I don't know if it's being maintained.

fcr--:
This would fix #181

### DIFF
--- a/flask_rebar/swagger_generation/generator_utils.py
+++ b/flask_rebar/swagger_generation/generator_utils.py
@@ -131,11 +131,6 @@ def _flatten(schema, definitions, base):
         for key, prop in properties.items():
             properties[key] = _flatten(schema=prop, definitions=definitions, base=base)
 
-        if sw.title in schema:
-            definitions_key = get_key(schema)
-            definitions[definitions_key] = schema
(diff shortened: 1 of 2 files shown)


fcr--:
All right, I added a test case and fixed the one with arrays.

fcr--:
@barakalon Would it be possible to do a release after merging this PR?

barakalon:
I don't have permissions to do a release.
Maybe @PratikPramanik does? Or @RookieRick ?

RookieRick:
@fcr-- I did a patch release last night that should include this (and a couple other recents PRs).  Let me know if any issues! Thanks!!

Issue #141, opened 2019-11-22
Allow hidden API


Sytten:
It would be great to be able to hide some APIs if they are not meant to be public. A parameter in the handles decorator would do the trick.

airstandley:
If we did this, I think we should also add a parameter to the generators. 
I'm not a huge fan of 'private' APIs in general, but I know that I would definitely want to be able to generate full documentation (including 'private' endpoints) for internal consumption.

mjanes:
Closed with https://github.com/plangrid/flask-rebar/pull/191

PR #191, opened 2020-07-16
Enabling hidden APIs in flask-rebar 

aliscott913:
https://jira.autodesk.com/browse/ACSBP-1520 #141
https://github.com/plangrid/flask-rebar/issues/141

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -239,6 +239,7 @@ class PathDefinition(
             "authenticators",
             "tags",
             "mimetype",
+            "hidden",
         ],
     )
 ):
@@ -327,6 +328,7 @@ class HandlerRegistry(object):
 
         self.prefix = normalize_prefix(prefix)
         self._paths = defaultdict(dict)
+        self._paths = defaultdict(dict)
         self.default_authenticators = default_authenticators
         self.default_headers_schema = default_headers_schema
         self.default_mimetype = default_mimetype
@@ -410,6 +412,7 @@ class HandlerRegistry(object):
                     authenticators=definition_.authenticators,
                     tags=definition_.tags,
                     mimetype=definition_.mimetype,
+                    hidden=definition_.hidden,
                 )
 
         return paths
@@ -435,6 +438,8 @@ class HandlerRegistry(object):
         authenticators=USE_DEFAULT,
         tags=None,
         mimetype=USE_DEFAULT,
+        hidden=False,
+
     ):
         """
         Registers a function as a request handler.
@@ -497,6 +502,7 @@ class HandlerRegistry(object):
             authenticators=authenticators,
             tags=tags,
             mimetype=mimetype,
+            hidden=hidden,
         )
 
     @deprecated_parameters(
@@ -519,6 +525,7 @@ class HandlerRegistry(object):
         authenticators=USE_DEFAULT,
         tags=None,
         mimetype=USE_DEFAULT,
+        hidden=False,
     ):
         """
         Same arguments as :meth:`HandlerRegistry.add_handler`, except this can
@@ -538,6 +545,7 @@ class HandlerRegistry(object):
                 authenticators=authenticators,
                 tags=tags,
                 mimetype=mimetype,
+                hidden=hidden,
             )
             return f
 
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -468,7 +468,7 @@ class HandlerRegistry(object):
         :param Type[USE_DEFAULT]|None|str mimetype:
             Content-Type header to add to the response schema
         :param bool hidden:
-            Used to control the generation of swagger documentation of APIs
+            if hidden, documentation is not created for this request handler by default
         """
         # Fix #115: if we were passed bare classes we'll go ahead and instantiate
         headers_schema = normalize_schema(headers_schema)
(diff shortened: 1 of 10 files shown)


flask_rebar/rebar.py
```diff
@@ -460,10 +463,12 @@ def add_handler(
             A list of authenticator objects to authenticate incoming requests.
             If left as USE_DEFAULT, the Rebar's default will be used.
             Set to None to make this an unauthenticated handler.
-        :param Sequence[str] tags:
+        :param Sequce[str] tags:
```
mjanes, line 466: Introduced a typo here

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -100,7 +102,6 @@ def generate(self, registry, host=None, sort_keys=True):
             default_security.extend(
                 self.authenticator_converter.get_security_requirements(authenticator)
             )
-
```
mjanes, line 103: Dropping this line looks accidental

RookieRick:
Couple of questions/thoughts:
1. Could/should we include this in the SwaggerV2Generator?  While the "@hidden" attribute is OpenApi 3 as I understand it, it doesn't look like there's anything that should prevent us from honoring the "hidden" setting in `PathDefinition` in the v2 generator.
2. Should prob add test coverage for this in e.g., https://github.com/plangrid/flask-rebar/blob/master/tests/swagger_generation/test_swagger_generator.py

mjanes:
@RookieRick are we maintaining Swagger 2 compatibility with the upgrade to flask-rebar 2.0?

RookieRick:
> @RookieRick are we maintaining Swagger 2 compatibility with the upgrade to flask-rebar 2.0?

Fantastic point LOL.  I guess we did kind of land on "maintain support for older versions of stuff only in 1.x branch" a while back.  Monday brain is mondaying.

Sytten:
My proposition is to keep the Swagger 2 compatibility as a plugin in Rebar V2. I know that some people still need it and are not ready to switch to OpenAPI 3.

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -175,6 +183,7 @@ def _get_paths(self, paths, default_headers_schema, default_security=None):
                 path_definition[sw.parameters] = path_params
 
             for method, d in methods.items():
+
```
mjanes, line 186: Nit, but can drop this line.

flask_rebar/rebar.py
```diff
@@ -464,6 +467,8 @@ def add_handler(
             Arbitrary strings to tag the handler with. These will translate to Swagger operation tags.
         :param Type[USE_DEFAULT]|None|str mimetype:
             Content-Type header to add to the response schema
+        :param bool hidden:
+            Used to control the generation of swagger documentation of APIs
```
mjanes, line 471: I would update this comment to indicate how behavior changed by this parameter. "if hidden, documentation is not created for this request handler by default" or similar

Sytten:
I will give it another look, but I will the plugin system is out of scope for this PR. It is already big enough to review.

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -139,6 +142,11 @@ def _get_paths(self, paths, default_headers_schema, default_security=None):
         for path, methods in paths.items():
             swagger_path, path_args = format_path_for_swagger(path)
 
+            if all(
+                d.hidden and not self.include_hidden for method, d in methods.items()
+            ):
```
RookieRick, line 147:
```suggestion
            if not self.include_hidden and all(
                d.hidden for method, d in methods.items()
            ):
```
Minor streamlining opportunity here since we only need to check `include_hidden` once

tests/swagger_generation/registries/hidden_api.py
```diff
@@ -0,0 +1,621 @@
...
+from flask_rebar import HeaderApiKeyAuthenticator
+from flask_rebar import compat
+from flask_rebar.swagger_generation import SwaggerV2Generator, SwaggerV3Generator
+
+rebar = Rebar()
+registry = rebar.create_handler_registry()
+
+swagger_v2_generator = SwaggerV2Generator()
+swagger_v3_generator = SwaggerV3Generator(include_hidden=True)
+hidden_swagger_v3_generator = SwaggerV3Generator()
```
RookieRick, line 13: Not a showstopper but I found the naming a little confusing and at first thought something was backwards in tests.  Would recommend calling this (the normal) one just `swagger_v3_generator` and the other something like `swagger_v3_generator_with_hidden` to make it explicit and obvious

tests/swagger_generation/registries/hidden_api.py
```diff
@@ -0,0 +1,621 @@
...
+            "type": "object",
+            "title": "Error",
+            "properties": {"message": {"type": "string"}, "errors": {"type": "object"}},
+            "required": ["message"],
+        },
+    },
+}
+
+
+EXPECTED_HIDDEN_SWAGGER_V3 = expected_swagger = {
```
RookieRick, line 251: Similar to above comment, I would consider giving this one the "normal" base name of `EXPECTED_SWAGGER_V3` and give the other a "specialized name".  Particularly since this is along side `EXPECTED_SWAGGER_V2`.

PR #200, opened 2020-07-28
chg: Parse Swagger path's summary from docstring.

jsonau:
The description will now be the content after the first blank line of a multi-line docstring.


docs/quickstart/swagger_generation.rst
```diff
@@ -73,10 +73,16 @@ The generator first checks for the value of ``endpoint`` specified when declarin
...
+^^^^^^^^^^^
+
+Swagger operations can have summaries. If a handler function has a docstring, the generator will use the content before the first blank line if any as the summary.
+
+
 description
 ^^^^^^^^^^^
 
-Swagger operations can have descriptions. If a handler function has a docstring, the generator will use this as a description.
+Swagger operations can have descriptions. If a handler function has a docstring with a blankline the generator will the texts after it as the description.
```
jab, line 85: This is missing a word between “will“ and “the texts”. Maybe “use” was intended?

Also, “blank line” is two words, not one, needs a comma after it, and the word “texts” should be singular here: “text”.

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -236,7 +238,12 @@ def _get_paths(self, paths, default_headers_schema, default_security=None):
                 }
 
                 if d.func.__doc__:
-                    path_definition[method_lower][sw.description] = d.func.__doc__
+                    docstring = chomp_right(chomp_left(d.func.__doc__))
```
airstandley, line 241: Do you mind explaining why `strip` is not sufficient here? It's not clear to me.

flask_rebar/swagger_generation/swagger_generator_v2.py
```diff
@@ -287,7 +287,12 @@ def _get_paths(self, paths, default_headers_schema, default_security=None):
                 }
 
                 if d.func.__doc__:
-                    path_definition[method_lower][sw.description] = d.func.__doc__
+                    docstring = d.func.__doc__.strip()
+                    doc_parts = str.split(docstring, "\n\n", 1)
+                    path_definition[method_lower][sw.summary] = doc_parts[0]
+                    path_definition[method_lower][sw.description] = (
+                        doc_parts[1].strip() if len(doc_parts) > 1 else ""
+                    )
```
airstandley, line 295: Might be worth moving to a shared function to avoid the repeated code?

flask_rebar/swagger_generation/swagger_generator_v2.py
```diff
@@ -287,7 +287,12 @@ def _get_paths(self, paths, default_headers_schema, default_security=None):
                 }
 
                 if d.func.__doc__:
-                    path_definition[method_lower][sw.description] = d.func.__doc__
+                    docstring = d.func.__doc__.strip()
+                    doc_parts = str.split(docstring, "\n\n", 1)
+                    path_definition[method_lower][sw.summary] = doc_parts[0]
```
airstandley, line 292: This already breaks backward compatibility, but still we may want to assign the docstring to the description when there is only 1 part instead of the summary so that it's a slightly less breaking change.

tests/swagger_generation/registries/multiple_authenticators.py
```diff
@@ -220,7 +220,11 @@ class NameAndOtherSchema(m.Schema):
     headers_schema=HeaderSchema(),
 )
 def get_foo(foo_uid):
-    """helpful description"""
+    """
+    a summary
+
+    helpful description
+    """
     pass
```
airstandley, line 228: We should make sure there is a test for the description only case, as well as the no docstring case.

Issue #187, opened 2020-06-30
SwaggerGenerator.register_authenticator_converter does not work.


airstandley:
The conversion of that method in #130 was incorrect and attempting to use the method now results in:

```
AttributeError: type object 'authenticator_converter_registry' has no attribute 'register_type'
```

Since we should drop that method for 2.0 I'm leaving this here as a note that we should change the documentation to drop all references to that method.

To create a proper converter:
```
from flask_rebar.swagger_generation import swagger_words as sw
from flask_rebar.swagger_generation.authenticator_to_swagger import AuthenticatorConverter

class MyAuthConverter(AuthenticatorConverter):
  def get_security_schemes(self, obj, context):
    return {
            obj.name: {sw.type_: sw.api_key, sw.in_: sw.header, sw.name: obj.header}
        }
  def get_security_requirements(self, obj, context):
    return [{obj.name: []}]

auth_converter = MyAuthConverter()
```

To convert an old style function into a new style converter:
```
from flask_rebar.swagger_generation.authenticator_to_swagger import make_class_from_method

from my_custom_stuff import MyAuthenticator

def my_conversion_function(authenticator):
    return {
        "name": MyAuthenticator._HEADER_NAME,
        "type": "apiKey",
        "in": "header"
    }

auth_converter = make_class_from_method(MyAuthenticator, my_conversion_function)
```

If you need to register a AuthenticatorConverter you can either
1) Instantiate your own registry and pass that in when instantiating the Generator.
```
from flask_rebar import SwaggerV3Generator
from flask_rebar.swagger_generation.authenticator_to_swagger import AuthenticatorConverterRegistry
from my_custom_stuff import auth_converter

my_auth_registry = AuthenticatorConverterRegistry()
my_auth_registry.register_type(auth_converter)

generator = SwaggerV3Generator(authenticator_converter_registry=my_auth_registry)
```

2) Register your converter with the global default registry.
```
flask_rebar.swagger_generation.authenticator_to_swagger import authenticator_converter_registry as global_authenticator_converter_registry
from my_custom_stuff import auth_converter

global_authenticator_converter_registry.register_type(auth_converter)

PR #202, opened 2020-08-04
authenticator_converter_registry is missing register_type

brockhaywood:
this targets the v1 branch since its against a deprecated function

SwaggerGenerator.register_authenticator_converter is currently crashing because register_type doesnt exist


airstandley:
https://github.com/plangrid/flask-rebar/issues/187

PR #205, opened 2020-08-07
Dropping python 2x and 3.4 

aliscott913:
JIRA Tickets | 
 --------------| 
 [ACSBP-1523](https://jira.autodesk.com/browse/ACSBP-1523)|


.github/workflows/pullrequests.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         python:
+
```
mjanes, line 15: Missed dropping 2.7 here

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -6,6 +6,7 @@
 
     :copyright: Copyright 2019 PlanGrid, Inc., see AUTHORS.
     :license: MIT, see LICENSE for details.
+
```
mjanes, line 9: Nit: no need for this change

PR #211, opened 2020-09-09
Modify validation to occur after dumping the object a serializable form

brockhaywood:
I'd like to also set a global or something to indicate if validation should occur and then pass that to marshall.

Maybe it should be set on a per handler registry basis and overridable per handler?

### DIFF
--- a/.github/workflows/pullrequests.yml
+++ b/.github/workflows/pullrequests.yml
@@ -52,7 +52,7 @@ jobs:
     - name: "Install black"
       run: |
         python -m pip install -U pip
-        python -m pip install black
+        python -m pip install black==19.10b0
         python -m pip freeze
     - name: Run black
       run: |
--- a/flask_rebar/compat.py
+++ b/flask_rebar/compat.py
@@ -17,8 +17,14 @@ def load(schema, data):
 
 
 def dump(schema, data):
-    obj = schema.load(data)  # Deserialize to trigger validation
-    return schema.dump(obj)
+    try:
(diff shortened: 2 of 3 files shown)


airstandley (approved):
LGTM
Though I think we'll want to bubble up the `validate` option to be something you can set/configure on the `rebar` object.

brockhaywood:
> LGTM
> Though I think we'll want to bubble up the `validate` option to be something you can set/configure on the `rebar` object.

Agreed. Thinking about tackling that in a follow up PR since this is what's needed to unblock further testing of rebar 2.X

airstandley:
> Maybe it should be set on a per handler registry basis and overridable per handler?

Sounds good. Though I'm not sure if it needs to be overridable per handler, or just overridable globally.

Something like:
global configuration can be 'validate', 'don't validate', or 'default -> uses handler setting'
handler configuration can be 'validate', 'don't validate', default is 'validate' to preserver existing behaviour

brockhaywood:
> > Maybe it should be set on a per handler registry basis and overridable per handler?
> 
> Sounds good. Though I'm not sure if it needs to be overridable per handler, or just overridable globally.
> 
> Something like:
> global configuration can be 'validate', 'don't validate', or 'default -> uses handler setting'
> handler configuration can be 'validate', 'don't validate', default is 'validate' to preserve existing behaviour

yeah, im not sure either. i was thinking similar to default headers checking behavior which is configurable globally and per handler

PR #216, opened 2020-10-26
fix: remove deprecated method and update docs

### LINKED ISSUE #187: SwaggerGenerator.register_authenticator_converter does not work.

RookieRick:
Resolves #187


docs/quickstart/swagger_generation.rst
```diff
@@ -169,35 +169,69 @@ Notice that since we've started to customize the swagger generator, we should sp
 Authenticators
 ^^^^^^^^^^^^^^
 
-We also need to tell the generator how to represent custom Authenticators as Swagger.
+We also need to tell the generator how to represent custom Authenticators as Swagger. Note that this changed somewhat in Flask-rebar 2.0:
```
airstandley, line 172: How about using similar syntax to the python docs where we add a note denoting if a feature had a breaking changed in a major version; could be a bit more concise?
* changed in 2.0
* new in 2.0

docs/quickstart/swagger_generation.rst
```diff
@@ -169,35 +169,69 @@ Notice that since we've started to customize the swagger generator, we should sp
...
 
-   class EasyGoingAuthenticator(Authenticator):
-       def authenticate(self):
-           pass
+    class MyAuthConverter(AuthenticatorConverter):
+        def get_security_schemes(self, obj, context):
+            return {
+                obj.name: {sw.type_: sw.api_key, sw.in_: sw.header, sw.name: obj.header}
+            }
+    def get_security_requirements(self, obj, context):
```
airstandley, line 186: under indented

mjanes (approved):

docs/quickstart/swagger_generation.rst
```diff
@@ -169,35 +169,69 @@ Notice that since we've started to customize the swagger generator, we should sp
...
-   from flask_rebar.authenticators import Authenticator
-   from flask_rebar import SwaggerV2Generator
-   from flask_rebar import Rebar
+    from flask_rebar.swagger_generation import swagger_words as sw
+    from flask_rebar.swagger_generation.authenticator_to_swagger import AuthenticatorConverter
 
-   class EasyGoingAuthenticator(Authenticator):
-       def authenticate(self):
-           pass
+    class MyAuthConverter(AuthenticatorConverter):
```
airstandley, line 181: Should include `AUTHENTICATOR_TYPE=MyAuthenticator`

tests/swagger_generation/test_generator_utils.py
```diff
@@ -217,49 +217,3 @@ def test_no_args(self):
...
-        pass
-
-
-def convert_my_customer_authenticator(authenticator):
-    definition = {"type": "basic"}
-    return "myAuth", definition
-
-
-class TestAuthenticatorConverter(unittest.TestCase):
-    def test_legacy_behaviour(self):
```
airstandley, line 233: I think this was the only AuthenticatorConverter test we may have had (My bad there).
Should we replace this with a test that check the new interface instead of just removing it?

airstandley (approved):
Found a couple typos and I have a few comments, but overall looks great!

docs/quickstart/swagger_generation.rst
RookieRick: I like it - will adjust

tests/swagger_generation/test_generator_utils.py
RookieRick: I looked around and thought I saw a couple of tests for the new approach but it's entirely possible I just glanced sideways at it ;) -- will take a better look and def agree if it's missing will add some tests

PR #233, opened 2021-05-10
Backport #230 to V1 branch

twosigmajab:
As per https://github.com/plangrid/flask-rebar/pull/230#issuecomment-834724160.

Quoting #230:
> Store "rebar" in app.extensions on rebar.init_app
> 
> Include handler_registries so that it's possible to enumerate all the
> Rebar APIs a Flask app provides given only a reference to the app object.

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -775,6 +775,8 @@ class Rebar(object):
         for registry in self.handler_registries:
             registry.register(app=app)
 
+        app.extensions["rebar"] = {"handler_registries": self.handler_registries}
+
     def _init_error_handling(self, app):
         @app.errorhandler(errors.HttpJsonError)
         def handle_http_error(error):


twosigmajab:
The only CI failure was the black formatting check, which failed due to existing code rather than any code introduced in this PR, so should probably be fixed separately (and perhaps need not block merging this).

RookieRick (approved):

RookieRick:
Thanks @twosigmajab !

RookieRick:
(I suppose that prob should have been a minor version bump and not a patch but 🤷 )

twosigmajab:
Looks like this is now in the [1.12.3 tag](https://github.com/plangrid/flask-rebar/commits/v1.12.3) and already
released to PyPI: https://pypi.org/project/flask-rebar/1.12.3/

Thanks @RookieRick! ❤️

Issue #232, opened 2021-04-30
Support for Flask v2


jab:
Flask v2 is due to come out in the next couple weeks, see https://palletsprojects.com/blog/end-python2-prereleases/

This issue tracks adding basic compatibility to Rebar with Flask v2 (making sure all current tests pass in CI and existing Rebar functionality works as expected). Taking advantage of Flask v2-specific features can be tracked in dedicated issues (e.g. #231).

dekim24:
working on this here https://github.com/plangrid/flask-rebar/pull/239

dekim24:
I merged the PR - can someone help bump version number / release the client? (cc @RookieRick or anyone who is an admin)

PR #239, opened 2021-06-10
add support for flask 2.0

dekim24:
- updated flask version requirements to allow flask 2.x
- `from flask import json_available` was deprecated in flask 1.x, and removed in flask 2.x. we should no longer need it since we don't support flask 0.x
- added a few version combinations for automated tests using flask v2. 
  - `flask 2` was released in May 2021 whereas `marshmallow 3` was released in 2019. based on that, I added a few older versions + the 3 most recent `marshmallow` versions to test alongside `flask 2.0`. 

in progress:

- install this package locally w/ flask v2 onto some internal repos that use flask-rebar, and see if there are any issues not identified through unit tests.
  - so far, I've tried updating `flask-toolbox` - a bunch of unit tests broke bc of changes in `marshmallow`, but no issues related to `flask-rebar` so far.


RookieRick (approved):
Fixes #232

PR #243, opened 2021-07-04
Add support for marshmallow-objects

### LINKED ISSUE #131: Integrate marshmallow-objects
One annoying thing about marshmallow is the fast that you get a dictionary instead of an object with attributes (I am bit jealous of [pydantic](https://github.com/samuelcolvin/pydantic)). We started using [marshmallow-objects](https://github.com/SVilgelm/marshmallow-objects) in our project. Since we already define some custom schemas, it would be very valuable to either use this lib or make something similar. Especially since we are going toward a v2 with some breaking changes anyway.

RookieRick:
Fixes #131

### DIFF
--- a/flask_rebar/utils/marshmallow_objects_helpers.py
+++ b/flask_rebar/utils/marshmallow_objects_helpers.py
@@ -1,12 +1,28 @@
-import marshmallow_objects as mo
+try:
+    import marshmallow_objects as mo
+
+    MARSHMALLOW_OBJECTS = True
+except ImportError:
+    MARSHMALLOW_OBJECTS = False
 
 
-class NestedTitledModel(mo.NestedModel):
-    """
-    Use this class instead of mashmallow_object.NestedModel if you need to supply
-    __swagger_title__ to override the default of {MyModelClass}Schema
-    """
+def get_marshmallow_objects_schema(model):
+    if MARSHMALLOW_OBJECTS and (
+        isinstance(model, mo.Model) or issubclass(model, mo.Model)
+    ):
+        return model.__get_schema_class__()
+    else:
+        return None
 
-    def __init__(self, nested, title, **kwargs):
-        super(NestedTitledModel, self).__init__(nested, **kwargs)
-        self.schema.__swagger_title__ = title
+
+if MARSHMALLOW_OBJECTS:
+
+    class NestedTitledModel(mo.NestedModel):
+        """
+        Use this class instead of mashmallow_object.NestedModel if you need to supply
+        __swagger_title__ to override the default of {MyModelClass}Schema
+        """
+
+        def __init__(self, nested, title, **kwargs):
+            super(NestedTitledModel, self).__init__(nested, **kwargs)
+            self.schema.__swagger_title__ = title
--- a/flask_rebar/utils/marshmallow_objects_helpers.py
+++ b/flask_rebar/utils/marshmallow_objects_helpers.py
@@ -1,0 +1,12 @@
+import marshmallow_objects as mo
+
+
+class NestedTitledModel(mo.NestedModel):
+    """
+    Use this class instead of mashmallow_object.NestedModel if you need to supply
+    __swagger_title__ to override the default of {MyModelClass}Schema
+    """
+
+    def __init__(self, nested, title, **kwargs):
+        super(NestedTitledModel, self).__init__(nested, **kwargs)
+        self.schema.__swagger_title__ = title
--- a/flask_rebar/utils/marshmallow_objects_helpers.py
+++ b/flask_rebar/utils/marshmallow_objects_helpers.py
@@ -26,3 +26,16 @@ if MARSHMALLOW_OBJECTS:
         def __init__(self, nested, title, **kwargs):
             super(NestedTitledModel, self).__init__(nested, **kwargs)
             self.schema.__swagger_title__ = title
+
+
+else:
+
+    class NestedTitledModel(object):
+        """
+        This version of NestedTitledModel will exist if marshmallow-objects is not present
+        """
+
+        def __init__(self):
+            raise ImportError(
+                "To use NestedTitledModel you must install marshmallow-objects"
+            )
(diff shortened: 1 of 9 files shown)


flask_rebar/utils/request_utils.py
```diff
@@ -94,9 +94,22 @@ def normalize_schema(schema):
     """
     This allows for either an instance of a marshmallow.Schema or the class
     itself to be passed to functions.
+    For Marshmallow-objects support, if a Model class is passed, return its __schema__
+    Tacit assumption, if __schema__ exists, we're dealing with marshmallow_objects, caveat emptor
```
RookieRick, line 98: The assumption here is kinda hack-ish, but my goal was to not introduce a hard dependency on marshmallow-objects. I went on the assumption that there is not likely to be a future version of marshmallow-objects that doesn't have `__schema__` - obvs not bulletproof but seems reasonable enough for this "bonus feature" - if a future version of marshmallow-objects were to break this, it would be trivial enough for implementing apps to pin `marshmallow-objects` accordingly (and at that point we can just note explicitly in docs that it doesn't work with those versions)

RookieRick:
So, I've found (so far) one major issue with the approach I tried here.  With `Schema` you can pass either the type or an instance when registering handlers (and if you pass a type, we instantiate it in `normalize_schema`. This won't work with marshmallow-objects `Model` because instantiating the model can't be done without supplying things like required fields. 
One workaround I'm toying with: 
1. Only (formally) support passing `Model` **_types_** to e.g., `@registry.handles`, and then
2. In `normalize_schema` make use of `Model` internal `__get_schema_class__` .... which, now that I type it "out loud", I realize is going to have unwanted side-effect of making `Model` behave just like `Schema` 😬  so that won't really work unless we bake in some additional logic that I suspect may ripple pretty far through the codebase.

Figured it might be prudent while I fiddle with bullet 2 to figure out just how ugly it may or may not be, to see if @Sytten @barakalon @airstandley or anybody else more familiar with `marshmallow-objects` can think of a less thorny approach ;)

tests/swagger_generation/test_swagger_generator.py
```diff
@@ -238,6 +239,16 @@ def update_foo(foo_uid):
             multiple_authenticators.swagger_v3_generator,
             multiple_authenticators.EXPECTED_SWAGGER_V3,
         ),
+        (
```
RookieRick, line 242: Bless you @barakalon for setting these tests up in such a delightfully extensible way 🏆  😃

barakalon: One gold star for me!

flask_rebar/utils/request_utils.py
barakalon: I think it's also possible to use some import logic here:

```python
try:
    import marshmallow_objects
    def get_marshmallow_objects_schema(...):
        if isinstance(schema, marshmallow_objects.Model):
            ...
except ImportError:
    def get_marshmallow_objects_schema(...):
        ...
```

A bit hacky... but that gives you a more explicit handle on types. 
🤷  just an idea. Both approaches seem equally messy. Unfortunately, we didn't make flask-rebar very pluggable...

gtmanfred: you could hide those functions a little lower down, so that the def statements dont have to be indented, and then just handle trying to import two different versions of that get function as well.

gtmanfred: and then try to import the marshmallow-objects version, and if you get an ImportError from that since it imports marshmallow_objects, then import the current version.

barakalon:
> In normalize_schema make use of Model internal __get_schema_class__ .... which, now that I type it "out loud", I realize is going to have unwanted side-effect of making Model behave just like Schema 😬 so that won't really work unless we bake in some additional logic that I suspect may ripple pretty far through the codebase.

Yeah, I think the tricky part here is loading the data _into_ the marshmallow model and exposing that object view functions. Otherwise, we're just extracting the schema from the models and still exposing plain ol' dictionaries.

flask_rebar/utils/request_utils.py
RookieRick: Great suggestions; I definitely like the idea of making the test for `marshmallow_objects` more explicit - will play around with this a bit before I finish off this PR.

RookieRick: Updated - note that I still have to do *some* odd nesting (class def in if statement) as I added a helper class that inherits from `marshmallow_objects.NestedModel` but it does feel cleaner as that is all now contained in `marshmallow_objects_helpers`

airstandley (approved):
lgtm

flask_rebar/utils/marshmallow_objects_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+def get_marshmallow_objects_schema(model):
+    if MARSHMALLOW_OBJECTS and (
+        isinstance(model, mo.Model) or issubclass(model, mo.Model)
+    ):
+        return model.__get_schema_class__()
+    else:
+        return None
+
+
+if MARSHMALLOW_OBJECTS:
```
airstandley, line 18: I'm thinking though how this will behave if MARSHMALLOW_OBJECTS is False and I think it might be confusing.

If I try "from marshmallow_objects_helpers import NestedTitledModel" and marshmallow_objects is not installed, then I'll get an import error along the line of 'can't import name "NestedTitleModel" from marshmallow_objects_helpers'. 

It might be better to have NestedTitleModel defined, but have it throw an error on instantiation?

RookieRick: I like that - I considered just creating a NOOP class but thought "eeh I don't want it to fail quietly" - will add something that fails loudly with a clear error message

tests/swagger_generation/test_swagger_generator.py
RookieRick: image
You own a significant share of these too! 😁

PR #245, opened 2021-07-23
chg: Add internal error codes 

### LINKED ISSUE #79: Customize error returned in Flask-rebar
Today we discussed creating our own class of errors because we wanted to have uniforms errors with a machine parsable `type`. Though we could add it just fine with the additional_data field, it is a bit painful to enforce it on every error thrown. The problem is that we can't currently modify the default errors that Flask-Rebar raise in the case of an invalid schema for example. I don't have a clear answer as to how we should handle this, but I figured I could ask if someone had an idea as to how we could do this.

RookieRick:
Fixes #79 - kinda..
Adds an internal `rebar_error_code` to those errors where we have different error messages for the same http status code.
Was also going to include an "error registry" but feeling like that's overkill (ref discussion on the linked Issue)
Note as part of 2.0 release, this is a breaking change.

### DIFF
--- a/flask_rebar/messages.py
+++ b/flask_rebar/messages.py
@@ -8,7 +8,6 @@
     :license: MIT, see LICENSE for details.
 """
 from collections import namedtuple
-from enum import IntEnum
 
 
 # machine-friendly equivalents of associated human-friendly messages
--- a/flask_rebar/messages.py
+++ b/flask_rebar/messages.py
@@ -11,19 +11,20 @@ from collections import namedtuple
 from enum import IntEnum
 
 
-class ErrorCode(IntEnum):
-    BODY_VALIDATION_FAILED = 1
-    EMPTY_JSON_BODY = 2
-    INTERNAL_SERVER_ERROR = 3
-    INVALID_AUTH_TOKEN = 4
-    INVALID_JSON = 5
-    MISSING_AUTH_TOKEN = 6
-    QUERY_STRING_VALIDATION_FAILED = 7
-    UNSUPPORTED_CONTENT_TYPE = 8
-    HEADER_VALIDATION_FAILED = 9
-    REQUIRED_FIELD_MISSING = 10
-    REQUIRED_FIELD_EMPTY = 11
-    UNSUPPORTED_FIELDS = 12
+# machine-friendly equivalents of associated human-friendly messages
+class ErrorCode:
+    BODY_VALIDATION_FAILED = "body_validation_failed"
+    EMPTY_JSON_BODY = "empty_json_body"
+    INTERNAL_SERVER_ERROR = "internal_server_error"
+    INVALID_AUTH_TOKEN = "invalid_auth_token"
+    INVALID_JSON = "invalid_json"
+    MISSING_AUTH_TOKEN = "missing_auth_token"
+    QUERY_STRING_VALIDATION_FAILED = "query_string_validation_failed"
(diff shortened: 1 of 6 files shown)


flask_rebar/messages.py
```diff
@@ -7,35 +7,80 @@
...
     :license: MIT, see LICENSE for details.
 """
+from collections import namedtuple
+from enum import IntEnum
 
-body_validation_failed = "JSON body parameters are invalid."
 
-empty_json_body = "Fields must be in JSON body."
+class ErrorCode(IntEnum):
+    BODY_VALIDATION_FAILED = 1
```
barakalon, line 15: Thoughts on using strings for error codes?

E.g.:
```python
class ErrorCode(IntEnum):
    BODY_VALIDATION_FAILED = "body_validation_failed"
    EMPTY_JSON_BODY = "empty_json_body"
```

I imagine these are mostly intended to be used by clients that want to programmatically respond to errors (e.g. render a translated error message on a UI somewhere), so I don't think the value matters that much, as long as it's unique. Might as well make it a little more human friendly... might help clients avoid some simple coding errors.

flask_rebar/rebar.py
```diff
@@ -859,9 +859,17 @@ def _create_json_error_response(
...
+            message_text = message.message
+            rebar_error_code = message.rebar_error_code
+        else:
+            message_text = message
+            rebar_error_code = None
+        body = {"message": message_text}
         if additional_data:
             body.update(additional_data)
+        if rebar_error_code:
+            body["rebar_error_code"] = rebar_error_code
```
barakalon, line 872: Perhaps we should set this value as an attribute of `Rebar`? e.g.

```python
class Rebar:
    def __init__(self, ...):
        self.error_code_attr = "rebar_error_code"

    ...

        body[self.error_code_attr] = rebar_error_code
```

It's a little clunky, but it gives an escape hatch for anyone who might not want to expose framework details in their API.

barakalon (approved):
Sweet!

flask_rebar/messages.py
RookieRick: Great point - will adjust

flask_rebar/rebar.py
RookieRick: Just to make sure I'm tracking correctly, let me try and re-state how I interpret this would work..  
For the "escape hatch," idea is they could explicitly set (given `rebar`, and instance of `Rebar`) `rebar.error_code_attr = None` and if we detect that where we would set `body[self.error_code_attr]` we just forego including that code in the response. 
Does that sound like what you had in mind, @barakalon ?

barakalon: Ah, I was more imagining...

```python
rebar.error_code_attr = "my_application_error_code"
# or...
rebar.error_code_attr = "error_code"
```

Ah, I hadn't thought about the case for omitting the error code entirely.

RookieRick: I think we can do both pretty easily 👍

RookieRick: Confirmed: easy peasy :) https://github.com/plangrid/flask-rebar/pull/245/commits/d45e1e716ee597149bdf514b9720a5274d0f2e7c

jordannielson (approved):

flask_rebar/rebar.py
barakalon: Sweet!

PR #253, opened 2021-08-04
fix: Remove now-pointless (and worse, broken) DisallowExtraFieldsMixin

### LINKED ISSUE #249: v2 Regression: DisallowExtraFieldsMixin broken
...as of the move to Marshmallow v3:

```python
>>> from flask_rebar.validation import DisallowExtraFieldsMixin
>>> print(DisallowExtraFieldsMixin.__doc__)  # this is no longer true:
    By default, Marshmallow will silently ignore fields that aren't included in a schema
    when serializing/deserializing.
    ...
```

ref: https://marshmallow.readthedocs.io/en/stable/upgrading.html#schemas-raise-validationerror-when-deserializing-data-with-unknown-keys
> Marshmallow 3.x schemas can deal with unknown keys in three different ways, configurable with the `unknown` option:
> EXCLUDE: drop those keys (same as marshmallow 2)
> INCLUDE: pass those keys/values as is, with no validation performed
> RAISE (default): raise a ValidationError

RookieRick:
Fixes #249 
With update from Marshmallow 2 to 3, the default behavior has changed - previously it was equivalent to `EXCLUDE` (ignore unexpected fields), whereas now it is `RAISE` (the behavior we previously exposed in our mixin).

Removing this mixin also required moving the appropriate "attribute setter" from a mixin-specific converter class to the primary `SchemaConverter` which has the side effect of making all of our generated OpenAPI specs EXPLICIT when it comes to `additionalProperties`, hence the massive proliferation of changes in tests

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -240,8 +240,12 @@ class SchemaConverter(MarshmallowConverter):
     def get_additional_properties(self, obj, context):
         if obj.unknown in (m.RAISE, m.EXCLUDE):
             return False
-        else:
+        elif obj.unknown is m.INCLUDE:
             return True
+        else:
+            raise ValueError(
+                f"Unexpected Schema.unknown value {obj.unknown} for {obj} "
+            )
 
 
 class FieldConverter(MarshmallowConverter):
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -25,7 +25,6 @@ from marshmallow.validate import Validator
 from flask_rebar import compat
 from flask_rebar.validation import QueryParamList
 from flask_rebar.validation import CommaSeparatedList
-from flask_rebar.validation import DisallowExtraFieldsMixin
 from flask_rebar.swagger_generation import swagger_words as sw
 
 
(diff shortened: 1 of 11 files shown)


jab (approved):
Thanks for fixing this @RookieRick!!

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -318,14 +328,6 @@ class ValidatorConverter(MarshmallowConverter):
     MARSHMALLOW_TYPE = Validator
 
 
-class DisallowExtraFieldsConverter(SchemaConverter):
-    MARSHMALLOW_TYPE = DisallowExtraFieldsMixin
-
-    @sets_swagger_attr(sw.additional_properties)
-    def get_additional_properties(self, obj, context):
-        return False
```
gtmanfred, line 326: When was this added, and did it ever work?

If it has ever actually done anything, we might need to make it a noop, and just make sure not to break peoples usage of the class and cause them to refactor it on a patch version bump.

gtmanfred (approved):
Just a couple questions on deleting classes before a patch release.

If you are confident that no one will have to change code because of this change, then I am ok with removing it.

flask_rebar/swagger_generation/marshmallow_to_swagger.py
RookieRick: `DisallowExtraFieldsMixin` has been around for roughly forever, it worked with Rebar 1.x.
It *should* have been removed before we released 2.0.0 with the change from Marshmallow 2 to 3.
It has never actually functioned in "2.x", which was why I went the "sneak this in as 2.0.1 even though it's technically a breaking change" route.  If somebody had a reference to this class and upgraded to 2.0 their code would have already broken like @jab 's 😂

gtmanfred: ok, sounds good to me. nuke it

PR #256, opened 2021-10-04
Add summary/description and example documentation features

shanempope:


gtmanfred:
@RookieRick do you remember what feedback we had for this? because it looks like the description is now becoming the summary?

marxav0319:
Tasks

- [X] Add `docstring_parser` to `extras_require`
- [X] Update to handle missing library `docstring_parser` gracefully
- [X] Add helper to add parsed descriptions to generated swagger
- [X] Add config to parse docstrings into long/short descriptions
- [] Add test coverage for different docstring edge-cases

flask_rebar/swagger_generation/generator_utils.py
```diff
@@ -282,3 +282,28 @@ def recursively_convert_dict_to_ordered_dict(obj):
         return [recursively_convert_dict_to_ordered_dict(item) for item in obj]
     else:
         return obj
+
+
+def get_parsed_method_docstring(method_definition, parse_docstring):
+    if method_definition.func.__doc__:
+        if parse_docstring:
+            try:
+                from docstring_parser import parse
```
gtmanfred, line 291: Can we import this at the top of the file like this 
```
try:
    from docstring_parser import parse
except ImportError:
    parser = None
```

And then do a check here for `if parser is not None`, This will prevent hitting this exception every time, and make this a little cleaner. Handling exceptions is also significantly slower than checking a boolean.

tests/helpers.py
```diff
@@ -14,3 +16,15 @@ def json(self):
...
+@pytest.fixture
+def docstring_parser_not_installed(monkeypatch):
+    original_imports = builtins.__import__
+
+    def mocked_imports(name, *args, **kwargs):
+        if name == "docstring_parser":
+            raise ImportError()
+        return original_imports(name, *args, **kwargs)
+
+    monkeypatch.setattr(builtins, "__import__", mocked_imports)
```
gtmanfred, line 30: The above change will also allow you to set `parser` as None, and be able to test the non imported behavior.

flask_rebar/swagger_generation/generator_utils.py
marxav0319: updated

tests/helpers.py
marxav0319: fixed, I removed this helper and just used `patch` to modify `parser` for the test.

gtmanfred (approved):
Can we add documentation about the docstring parser usage somewhere so that people can find it just by reading that?

Also, i would like to make sure @airstandley has a chance to comment, since he raised some of the initial concerns with the previous pr.

setup.py
```diff
@@ -29,7 +29,11 @@
         license="MIT",
         packages=find_packages(exclude=("test*", "examples")),
         include_package_data=True,
-        extras_require={"dev": development, "enum": ["marshmallow-enum~=1.5"]},
+        extras_require={
+            "dev": development,
+            "enum": ["marshmallow-enum~=1.5"],
+            "doc_parser": ["docstring_parser~=0.13"],
```
airstandley, line 35: It might be better to name this extra after the feature that it is providing. Though I'm not sure what exactly we should call this feature.
```suggestion
            "docstring-metadata": ["docstring_parser~=0.13"],
```

tests/swagger_generation/test_generator_utils.py
```diff
@@ -210,3 +214,79 @@ def test_no_args(self):
 
         self.assertEqual(res, "/health")
         self.assertEqual(args, tuple())
+
+
+class TestAddDocStringToPathDefinition(unittest.TestCase):
```
airstandley, line 219: Love the test coverage.

airstandley (approved):
Thanks for the PR! Sorry for the delayed review. 

I think there are a few places where the comments or documentation could be a bit clearer, but tests and code look fantastic.

PR #262, opened 2022-01-09
Fix filter_dump_only for NoneType nested fields

mbierma:
`filter_dump_only` will erroneously throw a `TypeError: filter_dump_only doesn't understand data type <class 'NoneType'>` error if a nested field marked as `allow_none=True` is `None`

### DIFF
--- a/flask_rebar/validation.py
+++ b/flask_rebar/validation.py
@@ -71,7 +71,8 @@ def filter_dump_only(schema, data):
             [item.loadable for item in processed_items],
             [item.dump_only for item in processed_items],
         )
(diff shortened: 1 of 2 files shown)


jordannielson (approved):
Changes look good

RookieRick:
Thanks @mbierma !  Guessing you discovered this the hard way; will get a patch release out that includes this 👍

RookieRick:
2.1.1 is out with this fix 🎉

mbierma:
Wow that was incredibly quick - thanks!!

PR #263, opened 2022-01-22
Only use EnumConverter if marshmallow-enum present

RookieRick:
Took a whack at addressing feedback from @barakalon on #257 (in a separate branch just in case we hate what I did)

I'm not sure how I feel about things like conditionally defining classes (I guess it's technically no "worse" than conditional imports, but it still "feels" dirty lol) and having to enforce order within `pytest` but... it does seem to work 🤣

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -687,7 +687,7 @@ def _common_converters():
         UUIDConverter(),
         ConstantConverter(),
     ]
-    if EnumField:
+    if EnumConverter.MARSHMALLOW_TYPE is not None:
         converters.append(EnumConverter())
 
     return converters
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -28,9 +28,10 @@ from flask_rebar.validation import CommaSeparatedList
 from flask_rebar.swagger_generation import swagger_words as sw
 
 try:
-    from marshmallow_enum import EnumField
+    from marshmallow_enum import EnumField, LoadDumpOptions
 except ImportError:
     EnumField = None
+    LoadDumpOptions = None
 
 
 # Special value to signify that a JSONSchema field should be left unset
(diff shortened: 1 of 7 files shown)


tests/swagger_generation/test_optional_converters.py
```diff
@@ -0,0 +1,37 @@
...
+import unittest
+from importlib import reload
+
+
+# HAAAAACKS - using importlib.reload will invalidate pre-existing imports in other test modules,
+# even when imported "as" something else..  so we'll just use pytest-order to ensure this test always runs LAST.
+@pytest.mark.order(-1)
+class TestOptionalConverters(unittest.TestCase):
+    def test_optional_enum_converter(self):
+        import flask_rebar.swagger_generation.marshmallow_to_swagger as _m_to_s
```
RookieRick, line 12: Note: I had to import the entire module (vs just `ALL_CONVERTERS` in order to use `importlib.reload` - which was the only way I could find/think of to reload the module with `marshmallow-enum` stuff mocked out.
Definitely open to suggestions for less nasty ways to accomplish what I did here

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -632,16 +633,36 @@ def convert(self, obj, openapi_version=2):
         )
 
 
-class EnumConverter(FieldConverter):
-    MARSHMALLOW_TYPE = EnumField
+if EnumField:
```
gtmanfred, line 636: I don't think you need this, you should be able to define it no matter what.

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -674,76 +693,28 @@ def get_enum(self, obj, context):
         UUIDConverter(),
         ConstantConverter(),
     ]
+    if EnumConverter:
+        converters.append(EnumConverter())
```
gtmanfred, line 697: Instead just append it is the marshmallow type is not none for the enum converter

tests/swagger_generation/test_marshmallow_to_swagger.py
```diff
@@ -8,15 +8,19 @@
     :license: MIT, see LICENSE for details.
 """
 import enum
+import mock
```
gtmanfred, line 11: Stop it. Mock is in unittest. Stop using the old mock module.


```suggestion
from unittest import mock
```

RookieRick: :shame: 😆

setup.py
```diff
@@ -9,10 +9,10 @@
     "gitchangelog>=3.0.4,<4.0.0",
     "jsonschema==3.0.2",
     "marshmallow-objects~=2.3",
-    "mock==2.0.0",
     "parametrize==0.1.1",
     "pre-commit>=1.14.4",
-    "pytest==4.6.8",
+    "pytest==6.2.5",
+    "pytest-order==1.0.1",
```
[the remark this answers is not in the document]
RookieRick: I think that predates my involvement with Rebar so I'm speculating, but I imagine that was done out of fear of breaking changes (although assuming people follow the rules we *should* be able to pin to at least minor versions so I'll adjust these :) )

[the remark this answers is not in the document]
RookieRick: (to clarify, I'm not insane - I do realize *I* updated that pytest pin and added pytest-order; I was referring to the prior pattern of hard pinning pytest which I lazy-followed 😆 )  Went ahead and loosened those two, but saving any broad re-pinning for a separate issue/pr

gtmanfred:
This looks good to merge and then rereview my pr.

Issue #265, opened 2022-03-29
BUG: DEPRECATED WERKZEUG METHOD


pirrofra:
File "/usr/local/lib/python3.8/site-packages/flask_rebar/__init__.py", line 5, in <module>
    from flask_rebar.rebar import (
  File "/usr/local/lib/python3.8/site-packages/flask_rebar/rebar.py", line 26, in <module>
    from flask_rebar.authenticators import Authenticator
  File "/usr/local/lib/python3.8/site-packages/flask_rebar/authenticators/__init__.py", line 3, in <module>
    from flask_rebar.authenticators.header_api_key import HeaderApiKeyAuthenticator
  File "/usr/local/lib/python3.8/site-packages/flask_rebar/authenticators/header_api_key.py", line 11, in <module>
    from werkzeug.security import safe_str_cmp
ImportError: cannot import name 'safe_str_cmp' from 'werkzeug.security' (/usr/local/lib/python3.8/site-packages/werkzeug/security.py)

safe_str_cmp has been removed in the lastest version of werkzeug (https://werkzeug.palletsprojects.com/en/2.1.x/changes/)

staccastacca:
Up

RookieRick:
OOF, thanks for the report - will get a patch out ASAP

RookieRick:
Update: ran into some other dependency issues in the process; I need to think a bit and possibly consult with other team members in the morning so will plan to release patch tomorrow morning. In meantime, I'd suggest just pinning to werkzeug < 2.1.0 to avoid breaking change and we'll work to resolve so that won't be necessary. Thanks again, and stay tuned

RookieRick:
OK, thanks for your patience folks. 2.1.2 is out now and should address this (by pinning `werkzeug`).
Hot on its heels should be 2.2.0 in which I'll remove any uses of stuff that was removed in `werkzeug 2.1.0`

PR #266, opened 2022-03-30
Pin werkzeug to avoid breaking change

RookieRick:
Related Issue #265 
This is PR 1 of 2, which will be basis for a patch release to unbreak any existing apps that rely on semantic versioning.
(Then will do a 2nd PR to prep a minor version release in which we'll switch dependencies)


gtmanfred (approved):
something else weird is going on, but this seems fine.

RookieRick:
> something else weird is going on, but this seems fine.

Haha yeah I'm not sure what's driving the new test failure but will figure that out before I merge this one just to play it safe

gtmanfred (dismissed):

PR #273, opened 2022-06-01
Try to fix dumping with data_key

jonls:
The test case demonstrates the issue: When using `data_key` in the schema, `filter_dump_only()` fails with `KeyError` because it assumes that the keys in the output are the same as the fields in the schema.

I've tried to fix this issue by adding a dict that translates from the output keys back to the schema fields.


gtmanfred (approved):

gtmanfred:
I will work on getting a patch release out with this tomorrow, thanks @jonls

Issue #274, opened 2022-06-28
prefixes with periods (".") cause flask's blueprint to complain that blueprint names can't have a period in them


rishi-matik:
To replicate, pass `prefix="/api/1.0"` as a kwarg to rebar.create_handler_registry. When init_app is called, Flask's blueprint complains that `ValueError: 'name' may not contain a dot '.' character`.

The issue originates at https://github.com/plangrid/flask-rebar/blame/master/flask_rebar/rebar.py#L620, where the blueprint name is prefixed by the url prefix for namespacing reasons I assume. I think the work around is to remove periods as that appears to be the only disallowed character in a Blueprint name (https://github.com/pallets/flask/blame/main/src/flask/blueprints.py#L186).

rishi-matik:
This is in reference to the blueprints being created by flask-rebar at https://github.com/plangrid/flask-rebar/blob/master/flask_rebar/swagger_ui/blueprint.py#L32 where the name being passed is generated at https://github.com/plangrid/flask-rebar/blame/master/flask_rebar/rebar.py#L620 by prepending the prefix (in our case this was "/api/1.0").

gtmanfred:
ok, this makes sense.

gtmanfred:
fixed by #275

PR #275, opened 2022-06-28
fix:Remove periods from blueprint name prefixes

rishi-matik:
https://github.com/plangrid/flask-rebar/issues/274

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -617,7 +617,7 @@ class HandlerRegistry(object):
         blueprint_name = "swagger_ui"
 
         if self.prefix:
-            blueprint_name = self.prefix + blueprint_name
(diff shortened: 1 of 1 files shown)


flask_rebar/rebar.py
```diff
@@ -617,7 +617,7 @@ def _register_swagger_ui(self, app):
         blueprint_name = "swagger_ui"
 
         if self.prefix:
-            blueprint_name = self.prefix + blueprint_name
+            blueprint_name = self.prefix.replace(".", "") + blueprint_name
```
gtmanfred, line 620: can we replace the '.' with '_' so there is still at least some delimiter to show we removed the period?

Other than that, this seems fine.

rishi-matik: makes sense to me. Will do.

PR #276, opened 2022-07-19
remove deprecated version check

dekim24:
Originally I got a warning about `distutils.version` being deprecated. As I was fixing that, I realized we're checking for versions of flask < 1.0 which aren't supported at all anymore, so I just removed this code entirely.


nlawrence22 (approved):
Looks reasonable to me, confirmed required flask version is > 1.0 [in setup.py](https://github.com/plangrid/flask-rebar/blob/master/setup.py#L33).

Thanks for pulling out the dead code!

PR #279, opened 2022-10-06
add flake8 + run all pre-commit hooks on entire repo

dekim24:
- Add flake8
- Run all pre-commit hooks on entire repo
- Pin click version (if you accidentally install click 7 it errors out if you try to run black locally)

### DIFF
--- a/flask_rebar/testing/swagger_jsonschema.py
+++ b/flask_rebar/testing/swagger_jsonschema.py
@@ -1,6 +1,11 @@
 # This is a JSONSchema document that validates a Swagger spec
 # We can use this to quickly check if the swagger we are producing is valid
 
+LONG_DESCRIPTION = (
+    "A longer description of the API. Should be different from the title. "
+    + "GitHub Flavored Markdown is allowed."
+)
+
 SWAGGER_V2_JSONSCHEMA = {
     "title": "A JSON Schema for Swagger 2.0 API.",
     "id": "http://swagger.io/v2/schema.json#",
@@ -66,7 +71,7 @@ SWAGGER_V2_JSONSCHEMA = {
                 },
                 "description": {
                     "type": "string",
-                    "description": "A longer description of the API. Should be different from the title.  GitHub Flavored Markdown is allowed.",
+                    "description": LONG_DESCRIPTION,
                 },
                 "termsOfService": {
                     "type": "string",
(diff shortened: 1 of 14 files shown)


flask_rebar/validation.py
```diff
@@ -8,21 +8,12 @@
     :license: MIT, see LICENSE for details.
 """
 from collections import namedtuple
-
-try:
-    from collections.abc import Mapping
-except ImportError:
-    from collections import Mapping
+from collections.abc import Mapping
```
dekim24, line 11: I believe this try/except was only required for compatibility with python 2, which we don't support anymore

tests/swagger_generation/test_marshmallow_to_swagger.py
```diff
@@ -8,6 +8,7 @@
     :license: MIT, see LICENSE for details.
 """
 import enum
+import unittest
```
dekim24, line 11: see `__main__` statement where we call unittest

tests/swagger_generation/test_swagger_generator.py
```diff
@@ -141,7 +141,7 @@ class Error(m.Schema):
                 variables={
                     "username": ServerVariable(
                         default="demo",
-                        description="this value is assigned by the service provider, in this example `gigantic-server.com`",
+                        description="this value is assigned by the service provider: `gigantic-server.com`",
```
dekim24, line 144: string too long, but this is just for a test anyway

flask_rebar/testing/swagger_jsonschema.py
```diff
@@ -1,6 +1,11 @@
 # This is a JSONSchema document that validates a Swagger spec
 # We can use this to quickly check if the swagger we are producing is valid
 
+LONG_DESCRIPTION = (
+    "A longer description of the API. Should be different from the title. "
+    + "GitHub Flavored Markdown is allowed."
```
airstandley, line 6: I would cut out "API" to avoid confusion if we are going to reuse this description for multiple 'description' blocks.
```suggestion
    "A longer description. Should be different from the title. "
    + "GitHub Flavored Markdown is allowed."
```

tests/swagger_generation/test_marshmallow_to_swagger.py
airstandley: We should just remove the `__main__` statement

airstandley (commented):
Looks good, left a few comments.

airstandley (approved):
Thanks for the PR @dekim24, looks amazing!

PR #281, opened 2022-11-01
Add the ability to auto import handlers

gtmanfred:
Having to explicitly import modules so that their handlers get registered on the registry is tedious. You have to make sure you import any new file.

Generally, these handlers are organized by directory for each registry, if that is the case, we should just import all of the modules in that directory to have the handlers be registered.

I also redid the todo app to use this method.


flask_rebar/rebar.py
```diff
@@ -548,6 +554,12 @@ def register(self, app):
         self._register_swagger_ui(app=app)
 
     def _register_routes(self, app):
+        if self.handlers is not None:
+            handlers = self.handlers if isinstance(self.handlers, list) else [self.handlers]
```
airstandley, line 558: We should probably handle the string case in the `__init__`, that would simplify the logic anytime we use self.handlers as it will then always be `Optional[List[str]]`

gtmanfred: you know i did that the first time, and then got an error because `[None]` is not None, and i ended up with this. I have a better solution though.

airstandley (commented):
Bit of nitpicking on the TODO refactor.

I feel like we should be leaning more towards Flask folder structure patterns than ACC ones.
Despite the fact I heavily dislike Flask's folder structure suggestions, our main OSS audience are Flask users coming from that documentation. Refs:
- https://flask.palletsprojects.com/en/2.2.x/tutorial/layout/
- https://flask.palletsprojects.com/en/2.2.x/patterns/packages/

The handlers module and schemas.py are reasonable breaks from the Flask pattern that I'm happy for us to recommend.
I'm not as much a fan of using app_init.py + rebar.py  for the TODO app. I think we should be combining these into a single app.py file.

airstandley (approved):
Rebar registry handler module loading code looks good.

PR #283, opened 2023-02-20
[PJL-7600] fix minor bug in swagger v3 generator

dekim24:
I generated a swaggerv3 client of one of our repos, copy pasted it here https://editor.swagger.io/ and noticed there were some validation errors. The fix requires 2 steps:

1. `style` needs to be defined at the `parameter` level - move it out of the `schema` level (this was the root cause of the validation error). example error here: 

```
Structural error at paths./v1/<schema_name>
should NOT have additional properties
additionalProperty: style
```

2. after doing that, I got a different validation error because the value of `style: simple` was invalid. basically the way we want to handle a CommaSeparatedList is to format a *query param* like this: `filter[name]=abc123,def567`. the correct style we need to use here is `style: form, explode: False` as discussed here https://swagger.io/docs/specification/serialization/ for query params. `style: simple` is actually for path parameters which would be something like `/v1/accounts/abc123,def567` which (as I understand it) is not the use case for `CommaSeparatedList`.

example error here:

```
Structural error at paths./v1/<parameter_name>.style
should be equal to one of the allowed values
allowedValues: form, spaceDelimited, pipeDelimited, deepObject
```

 JIRA Tickets | 
 --------------| 
 [PJL-7600](https://jira.autodesk.com/browse/PJL-7600)|

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -472,7 +472,11 @@ class CsvArrayConverter(ListConverter):
 
     @sets_swagger_attr(sw.style)
     def get_style(self, obj, context):
-        return sw.simple if context.openapi_version == 3 else UNSET
+        return sw.form if context.openapi_version == 3 else UNSET
+
+    @sets_swagger_attr(sw.explode)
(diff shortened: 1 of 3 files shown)


gtmanfred (approved):
Awesome, thanks for all the context.

PR #285, opened 2023-05-11
ACSBP3936 Add Enum converter for marshmallow.field.Enum

auramix:
Marshmallow 2 didn't have a built-in for Enums. This adds a swagger converter and tests.

We could also remove the old converter and cut a new major release, if desired.

 JIRA Tickets | 
 --------------| 
 [ACSBP-3936](https://jira.autodesk.com/browse/ACSBP-3936)|


gtmanfred:
The new enum field should be the same as the old enum field, but we can supersede it, So i opened up a pr to do similar if the newer enum field was available.
https://github.com/plangrid/flask-rebar/pull/286/files#diff-0248fbda166bd5a75d683d813429d908986ebe9146fac4699f4e2d19ebfbda72R31-R37

I also updated all the github actions stuff.

PR #286, opened 2023-05-11
update to use Enum in 3.18.0

gtmanfred:
Also update all the github actions to the latest versions and newer marshmallow versions for testing.


gtmanfred:
i dropped python 3.7 from the test suite. It should still pass here, it is just that the importlib.metadata stuff was added in 3.8 so the test suite wont finish, and 3.7 eols in just over a month.

flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -27,6 +27,13 @@
 from flask_rebar.validation import CommaSeparatedList
 from flask_rebar.swagger_generation import swagger_words as sw
 
+try:
+    EnumSymbol = m.fields.EnumSymbol
+    EnumValue = m.fields.EnumValue
```
auramix, line 32: Why are these fields not documented by marshmallow?

gtmanfred: Oh, i was right initially, they merged it after the fact https://github.com/marshmallow-code/marshmallow/commit/fa0ee03e82d2b73dc00bdeb75b266aba99be8d0b

.github/workflows/pullrequests.yml
```diff
@@ -12,41 +12,32 @@ jobs:
...
         - marshmallow=='3.15.*' flask=='2.1.*' werkzeug=='2.1.*'
+        - marshmallow=='3.19.*' flask=='2.1.*' werkzeug=='2.1.*'
+        - marshmallow=='3.0.*' flask=='2.2.*' werkzeug=='2.2.*'
+        - marshmallow=='3.10.*' flask=='2.2.*' werkzeug=='2.2.*'
+        - marshmallow=='3.15.*' flask=='2.2.*' werkzeug=='2.2.*'
+        - marshmallow=='3.19.*' flask=='2.2.*' werkzeug=='2.2.*'
+        - marshmallow=='3.0.*' flask=='2.3.*' werkzeug=='2.3.*'
+        - marshmallow=='3.10.*' flask=='2.3.*' werkzeug=='2.3.*'
+        - marshmallow=='3.15.*' flask=='2.3.*' werkzeug=='2.3.*'
+        - marshmallow=='3.19.*' flask=='2.3.*' werkzeug=='2.3.*'
```
airstandley, line 35: Looks like a lot of combinations got dropped from the test matrix. Was this change all due to dropping 3.7 support?

.github/workflows/pullrequests.yml
```diff
@@ -56,16 +47,22 @@ jobs:
...
     - name: Run Tests
       run: |
-        python -m pytest
+        python -m pytest -v -ra --junitxml=pytest.xml
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v2
+      if: always()
+      with:
+        report_paths: pytest.xml
+
```
airstandley, line 56: Why the new action to publish the junit report?

flask_rebar/swagger_generation/marshmallow_to_swagger.py
airstandley: Can you explain this a bit?
It doesn't seem to me like it will address the originally raise issue that the `Enum` field introduced in marshmallow 3.18 is not supported.

tests/swagger_generation/test_marshmallow_to_swagger.py
```diff
@@ -135,27 +136,49 @@ def test_primitive_types(self):
...
+        for field, result in [
+            (
+                me.EnumField(StopLight),
+                {"enum": ["green", "yellow", "red"], "type": "string"},
+            ),
+            (
+                me.EnumField(StopLight, by_value=True),
+                {"enum": [1, 2, 3], "type": "integer"},
+            ),
+        ]:
```
airstandley, line 157: Tests for `m.EnumField` are already encompassed in `test_primative_types`. Did you mean to test a different field?

.github/workflows/pullrequests.yml
gtmanfred: I dropped a lot because we have a limit on how many we can run in any given time, and i just don't know that arbitrarily testing all of them is useful. Especially when we are adding new flask and werkzeug versions.

.github/workflows/pullrequests.yml
gtmanfred: So that you don't have to scroll through the test output to view what had failed, and instead it shows it like a test report in the actions output.

tests/swagger_generation/test_marshmallow_to_swagger.py
gtmanfred: yes, i meant to switch these to EnumSymbol and EnumValue, but i need to redo this again, because they actually publish a regular Enum field on its own.

.github/workflows/pullrequests.yml
airstandley: All I'm seeing is a flood of `JUnit Test Report` actions which do not seem at all more understandable or parsable. What am I missing?

.github/workflows/pullrequests.yml
gtmanfred: I just picked round numbers for marshmallow. and then 3.19.x because it has the new enum field.

.github/workflows/pullrequests.yml
gtmanfred: That is because there are no failures, but here is the test report https://github.com/plangrid/flask-rebar/pull/286/checks?check_run_id=13404513126

they will show you what the failures are if there are any

here is an example
https://github.com/gtmanfred/teststack/runs/12086499726

.github/workflows/pullrequests.yml
gtmanfred: There is a maximum amount of matrix runs to 256.
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy

airstandley: Thinking about it; we should probably separate the Marshmallow versions from the Flask/Werkzuerg versions.

Looking back it seems like the following are the only versions where we should be checking against regressions
- Werzuerg 2.1 (dropped a method - https://github.com/plangrid/flask-rebar/pull/268)
- Marshmallow 3.18 (Introduces Enum)

Can we just re-write our matrix as
```
marshmallow:
 - 3.17.*
 - 3.18.*
 - 3.19.*
 flask:
 - flask=='2.0.*' werkzeug=='2.0.*'
 - flask=='2.1.*' werkzeug=='2.1.*'
 - flask=='2.3.*' werkzeug=='2.3.*'
```

.github/workflows/pullrequests.yml
airstandley: Thanks for the examples.
Is there any way to configure it so that the output show what combination was used on that test run? As it stands I personally just find that view confusing as it's not clear to me which entries in the test matrix experienced which failures.

.github/workflows/pullrequests.yml
gtmanfred: looks like it worked.

.github/workflows/pullrequests.yml
gtmanfred: I don't think you can directly tie the output to the job that ran, but all it is doing is dumping the output into the job summary, and from there you can see which job had failed tests, and that can be used to take you to the full log.
https://github.com/plangrid/flask-rebar/actions/runs/4951665623

gtmanfred: Screenshot 2023-05-11 at 3 20 00 PM
You would do that by clicking "View job logs"

airstandley: Thanks, appreciate all the answers. I'm sold 👍

flask_rebar/swagger_generation/marshmallow_to_swagger.py
airstandley, line 1: This is way more elegant that I expected this fix to be. Awesome!

airstandley (approved):
🔥

flask_rebar/swagger_generation/marshmallow_to_swagger.py
gtmanfred: I think i thought of a way to shove the matrix information into the test report output.

PR #287, opened 2023-05-11
Bump version: 2.3.0 → 2.4.0

auramix:


setup.py
```diff
@@ -4,7 +4,7 @@
 
 # packages required for local development and testing
 development = [
-    "black==22.3.0",
+    "black==22.4.0",
```
gtmanfred, line 7: we need to update the bumpversion.cfg file to only search for `version="{current_version}",`

auramix: Yes

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.3.0
+current_version = 2.4.0
 commit = True
 tag = True
```
gtmanfred, line 4:
```
[bumpversion]
current_version = 2.4.0
commit = True
tag = True

[bumpversion:file:setup.py]
search = version="{current_version}",
replace = version="{new_version}",
```

PR #289, opened 2023-05-16
Update tag.yml

auramix:


gtmanfred (approved):

.github/workflows/tag.yml
```diff
@@ -8,13 +8,13 @@ on:
 jobs:
   build-n-publish:
     name: Build and publish Python 🐍 distributions 📦 to PyPI
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
nlawrence22, line 11: Just confirming we want to be on a rolling release here, and not pinned to a specific version?  

I'm ok with this, and we should definitely update our pin from 18.04 regardless, but just making sure we're comfortable with the tradeoffs of not pinning a specific version.

nlawrence22 (approved):

PR #292, opened 2023-07-21
install mypy + update dev deps + update pre commit + begin type hinting

dekim24:
This is basically the minimum required code in order to get `mypy .` to pass. In future PR(s) I will work through untyped definitions and add type hints to everything, with the end goal being to set `check_untyped_defs = True` in the config for the entire repo.

Also I was getting an error because the pinned version of jsonschema doesn't work with python 3.11 when I run tests on my laptop (it seems to work OK in CI/CD for some reason). I went ahead and updated all dev deps to latest possible version to get past this.

Also ran `pyupgrade` to remove old compatibility code since we now require python3.8+


flask_rebar/__init__.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import unicode_literals
```
dekim24, line 1: I saw that we only run CI/CD unit tests for python 3.8 and later. I assumed this meant we don't support python 2.0 anymore, but maybe that isnt the case?

or maybe we should release a new major version that drops python 2.0 support?

flask_rebar/rebar.py
```diff
@@ -35,12 +34,6 @@
 from flask_rebar.swagger_generation import SwaggerV2Generator
 from flask_rebar.swagger_ui import create_swagger_ui_blueprint
 
-# Deal with maintaining (for now at least) support for 2.7+:
-try:
```
dekim24, line 39: same question as above

flask_rebar/swagger_generation/authenticator_to_swagger.py
```diff
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from typing import Type
```
dekim24, line 2: this import is required for python <= 3.8, 3.9 and later support `type[]`

flask_rebar/__init__.py
airstandley: We are currently supporting Python 3.8 through Python 3.11 : https://github.com/plangrid/flask-rebar/blob/master/setup.py#L44-L47

airstandley: I for one would welcome a PR that cleans up all the compatability cruft that we no longer need to maintain 👍

dekim24: I ran across a pre-commit hook that did this automatically, let me see if I can apply that here

dekim24: @airstandley I ran `pyupgrade` on the entire repo to remove pre-3.8 compatibility code

mypy.ini
airstandley, line 1: We should put these in the [pyproject.toml](https://github.com/plangrid/flask-rebar/blob/master/pyproject.toml) along with the black configurations.

airstandley (changes requested):
Looks great!
I have some minor nitpicks that I'd love to see addressed.

airstandley (approved):

PR #295, opened 2023-08-14
allow specifying swagger documents for path converters

gtmanfred:
without much changing elsewhere, this allows us to set enum types for custom path converters in flask.

for example: `/v1/preferences/<preference_type:scope>` where scope could be `user`, `project`, or `account`. We can define an object with a `to_swagger` function as part of the [Flask Converter](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Flask.url_map) that is defined.

### DIFF
--- a/tests/swagger_generation/registries/legacy.py
+++ b/tests/swagger_generation/registries/legacy.py
@@ -12,13 +12,9 @@ registry = rebar.create_handler_registry()
 
 
 class UUIDPathConverter:
-
     @staticmethod
     def to_swagger():
-        return {
-            sw.type_: sw.string,
-            sw.format_: sw.uuid
-        }
+        return {sw.type_: sw.string, sw.format_: sw.uuid}
 
 
 swagger_v2_generator = SwaggerV2Generator()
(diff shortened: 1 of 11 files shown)


airstandley (commented):
Looks good, but would love to see a test and/or example of this in the docs.

gtmanfred:
ok @airstandley I have gotten around to adding documentation and tests :)

tests/swagger_generation/registries/legacy.py
```diff
@@ -5,12 +5,22 @@
 from flask_rebar import HeaderApiKeyAuthenticator
 from flask_rebar import compat
 from flask_rebar.swagger_generation import SwaggerV2Generator, SwaggerV3Generator
+from flask_rebar.swagger_generation import swagger_words as sw
 
 rebar = Rebar()
 registry = rebar.create_handler_registry()
 
+
+class UUIDPathConverter:
```
airstandley, line 14: Awesome

airstandley (approved):
🎉 Looks brilliant, thanks for adding the tests and docs!

PR #298, opened 2023-09-24
fix: Prevent hidden methods from being included in schema

mbierma:
**Summary of issue**:

APIs specified with `hidden=True` are included in the schema unless _all_ methods for the path are marked as `hidden=True`. For example, if we define two methods on the `/foos/<foo_uid>` path:

```python
@registry.handles(
    rule="/foos/<foo_uid>",
    method="PATCH",
)
def update_foo(foo_uid):
    pass

@registry.handles(
    rule="/foos/<foo_uid>",
    method="DELETE",
    hidden=True
)
def delete_foo(foo_uid):
    pass
```

The `DELETE` request will be included in the generated spec, due to the requirement for [all methods of the path](https://github.com/plangrid/flask-rebar/blob/6228a2410246e1a7e1f5feedb05b01fb591face0/flask_rebar/swagger_generation/swagger_generator_v3.py#L142-L145) to have `hidden=True` set.

**Summary of change**:

Update the generator to respect `hidden=True` on methods even if all methods for the path are not marked as `hidden=True`.

### DIFF
--- a/flask_rebar/swagger_generation/swagger_generator_v3.py
+++ b/flask_rebar/swagger_generation/swagger_generator_v3.py
@@ -180,6 +180,9 @@ class SwaggerV3Generator(SwaggerGenerator):
                 path_definition[sw.parameters] = path_params
 
             for method, d in methods.items():
+                if d.hidden:
+                    continue
+
                 responses_definition = {
                     sw.default: self._get_response_definition(
                         self.default_response_schema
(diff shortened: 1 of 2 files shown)


gtmanfred (approved):

gtmanfred:
Thanks for this. Can you run black on this, and then we should be able to merge it.

PR #299, opened 2023-09-26
feat: Support optional short summary for paths

mbierma:
**Summary of issue**:
In OpenAPI, paths may have an optional short summary and a longer description for documentation purposes. However, it is not possible to provide a summary when defining an endpoint.

**Summary of change**:
Support the ability to provide a short summary for an endpoint using the handles decorator:

```python
@registry.handles(
    rule="/foos", method="GET", summary="Foos"
)
```


gtmanfred (approved):
This seems fine to me, just needs the black tests to pass too.

PR #300, opened 2023-10-16
add mypy type hints to all non-test code + update mypy config

dekim24:
- Update mypy type hint to require type hints on non-test code
- run `mypy .` and address all type errors by adding type hints
- fix a few typos I discovered in the process

I will test this by installing this version of the library locally and make sure that type hinting passes on a few of our repos that use flask-rebar.
- [x] test library 1
- [x] test library 2


flask_rebar/rebar.py
```diff
@@ -8,21 +8,39 @@
     :copyright: Copyright 2018 PlanGrid, Inc., see AUTHORS.
     :license: MIT, see LICENSE for details.
 """
+from __future__ import annotations
```
dekim24, line 11: this is for type hints like `def copy(self) -> InstanceOfThisClass` where I'm referencing a class from inside it

flask_rebar/swagger_generation/generator_utils.py
```diff
@@ -10,13 +10,32 @@
...
+)
+
+from flask_rebar.authenticators import Authenticator
 from flask_rebar.utils.defaults import USE_DEFAULT
 from flask_rebar.swagger_generation import swagger_words as sw
 from flask_rebar.swagger_generation.marshmallow_to_swagger import get_swagger_title
+from marshmallow import Schema
+
+if TYPE_CHECKING:
+    from flask_rebar.rebar import HandlerRegistry
```
dekim24, line 34: rebar also imports from files in swagger_generation/ so this is needed to avoid circular imports

flask_rebar/validation.py
```diff
@@ -17,12 +18,14 @@
...
-def filter_dump_only(schema, data):
+def filter_dump_only(
+    schema: Schema, data: Union[Mapping[str, Any], List[Mapping[str, Any]]]
+) -> FilterResult:
     """
     Return a filtered copy of data in which any items matching a "dump_only" field are separated
     :param schema: Instance of a Schema class
-    :param data: Dict or collection of dicts with data
-    :return: Union[FilterResult, list[FilterResult]]
+    :param data: Mapping or List of Mappings with data
```
dekim24, line 27: think whether or not we use `Mapping / Dict` or `List/Collection` is actually dictated by the type hints in marshmallow

flask_rebar/swagger_generation/swagger_generator_v3.py
```diff
@@ -119,7 +140,7 @@ def generate(self, registry, host=None, sort_keys=True):
         if self.tags:
             swagger[sw.tags] = [tag.as_swagger() for tag in self.tags]
 
-        servers = self.servers or []
+        servers = list(self.servers or [])
```
dekim24, line 143: can't `append()` to a `Sequence`

flask_rebar/utils/request_utils.py
```diff
@@ -110,28 +142,30 @@ def normalize_schema(schema):
...
-        return
+        return errors.BadRequest(msg=msg)
 
     copied = copy.deepcopy(errs)
 
     _format_marshmallow_errors_for_response_in_place(copied)
 
     additional_data = {"errors": copied}
-    message, _ = msg if isinstance(msg, tuple) else msg, None
+    message, _ = msg if isinstance(msg, tuple) else (msg, None)
```
dekim24, line 163: I think this was a typo - the original code was basically equivalent to `message, _ = (msg if isinstance(msg, tuple) else msg), None` which is identical to `message = msg` and I don't think that was the intent.

flask_rebar/rebar.py
airstandley: Which Python versions is this needed for? (Is this only required for 3.8 support?)

tests/test_errors.py
```diff
@@ -193,6 +193,7 @@ def test_json_body_parameter_validation(self):
             path="/stuffs", data={"foo": "one", "bar": "not-an-email"}
         )
         expected = messages.body_validation_failed._asdict()
+        expected.pop("rebar_error_code")
```
airstandley, line 196: Why do we need to pop that key now?

airstandley (commented):
I have a couple of questions/concerns about some of the logic changes, but otherwise looks brilliant.

flask_rebar/swagger_generation/swagger_generator_v3.py
airstandley: It would probably be better to type `self.servers` as a list and convert the provided value in the `__init__` rather than creating a new list every time `generate` is called.

flask_rebar/rebar.py
dekim24: here it says TBD on when this will become mandatory https://docs.python.org/3/library/__future__.html

tests/test_errors.py
dekim24: I fixed this now, it was actually related to this typo https://github.com/plangrid/flask-rebar/pull/300#discussion_r1435785217 - I think we actually don't need that line of code at all, so that way it passes an ErrorMessage object (rather than the `message` which is just a string) and we get the full detail in the response including the `rebar_error_code`.

airstandley (approved):
lgtm, thank you for the contribution!

Issue #231, opened 2021-04-30
Support async request handlers


jab:
Flask v2 (due to come out in the next couple weeks) adds limited support for async request handlers, see e.g.:
- http://flask.palletsprojects.com/en/master/async-await/
- https://palletsprojects.com/blog/end-python2-prereleases/

As the docs above say:
> Existing Flask extensions only expect views to be synchronous. If they provide decorators to add functionality to views, those will probably not work with async views because they will not await the function or be awaitable. Other functions they provide will not be awaitable either and will probably be blocking if called within an async view.

This issue tracks adding support to Rebar specifically for Flask v2 async request handlers.

PR #311, opened 2024-04-03
Async handler support

jonls:
As far as I can tell, all that is needed is to wrap the call to the handler in `current_app.ensure_sync()` as described in https://flask.palletsprojects.com/en/2.3.x/async-await/#extensions

Issue for this feature request: #231


gtmanfred (approved):
i would love to see us eventually get to a place where we can also support Quart.

Issue #314, opened 2024-08-06
After upgrading from 2.4.1 to 3.3.0, generated swagger schemas are incorrect for `fields.List(fields.Nested(...), allow_none=True)`


kaiku:
### Problem

The latest version of flask-rebar generates a swagger.json that "expands" `$ref` to nested schemas specifically when using `fields.List(fields.Nested(...), allow_none=True)`.

### Details

I have marshmallow schemas with fields that look like this:
```
class NestedSchema(Schema):
    foo = fields.String()

class ParentSchema(Schema):
    data = fields.List(fields.Nested(NestedSchema), required=False, allow_none=True)
```

In 2.4.1, using `swagger_generator_v3` produces a swagger.json that looks like this:

```
"ParentSchema": {
  "additionalProperties": false,
  "properties": {
    "data": {
      "items": {
        "$ref": "#/components/schemas/NestedSchema"
      },
      "nullable": true,
      "type": "array"
    }
  },
  "title": "ParentSchema",
  "type": "object"
}
```

In 3.3.0, it looks like this:

```
"ParentSchema": {
  "additionalProperties": false,
  "properties": {
    "data": {
      "items": {
        "additionalProperties": false,
        "description": "A nested schema",
        "properties": {
          "foo": {
            "type": "string"
          },
        },
        "required": [
          "foo"
        ],
        "title": "NestedSchema",
        "type": "object"
      },
      "type": [
        "array",
        "null"
      ]
    }
  },
  "title": "ParentSchema",
  "type": "object"
}
```

**Expected:**

```
"ParentSchema": {
  "additionalProperties": false,
  "properties": {
    "data": {
      "items": {
        "$ref": "#/components/schemas/NestedSchema"
      },
        "required": [
          "foo"
        ],
        "title": "NestedSchema",
        "type": "object"
      },
      "type": [
        "array",
        "null"
      ]
    }
  },
  "title": "ParentSchema",
  "type": "object"
}
```

The important difference here is that instead of a `$ref`, the swagger codegen now expands `NestedSchema` into an embedded dict. From a raw API perspective there is no functional difference in the spec, but this does cause problems in our unit tests because we use a python client generated by swagger-codegen-cli.jar to make calls against our service, and responses from that _are_ different.

### Investigation
`allow_none=True` seems to be the culprit. If I remove that line, I get a nice ref again. However, I can't do this because it would break our API contract of allowing that field to actually be None.

[...]

PR #315, opened 2024-08-06
Fix schema flattener to support list-type `type` field introduced in OpenAPI 3.1

kaiku:
### Summary

OpenAPI 3.1 removed `nullable: true` and expanded `type` to be either a string or a list of strings. For nullable fields, `type` will be an array that has "null" as a value: `type: ["object", "null"]`.

This PR reproduces and fixes https://github.com/plangrid/flask-rebar/issues/314 by teaching `flatten` to expect either a string or list of strings, allowing it to correctly resolves nested objects into references.

### DIFF
--- a/flask_rebar/swagger_generation/generator_utils.py
+++ b/flask_rebar/swagger_generation/generator_utils.py
@@ -143,9 +143,9 @@ def _flatten(
     # With OpenAPI 3.1, this will be a list of allowed types that includes sw.null if the field is nullable.
     schema_type: str | list[str] | None = schema.get(sw.type_)
     schema_types = []
-    if type(schema_type) is str:
+    if isinstance(schema_type, str):
         schema_types = [schema_type]
-    elif type(schema_type) is list:
+    elif isinstance(schema_type, list):
         schema_types = schema_type
 
     subschema_keyword = _get_subschema_keyword(schema)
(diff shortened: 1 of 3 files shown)


VerinaG (approved):
LGTM

flask_rebar/swagger_generation/generator_utils.py
```diff
@@ -140,15 +140,22 @@ def flatten(schema: Dict[str, Any], base: str) -> Tuple[Dict[str, str], Dict[str
 def _flatten(
     schema: Dict[str, Any], definitions: Dict[str, Any], base: str
 ) -> Dict[str, str]:
-    schema_type = schema.get(sw.type_)
+    # With OpenAPI 3.1, this will be a list of allowed types that includes sw.null if the field is nullable.
+    schema_type: str | list[str] | None = schema.get(sw.type_)
+    schema_types = []
+    if type(schema_type) is str:
+        schema_types = [schema_type]
+    elif type(schema_type) is list:
```
gtmanfred, line 148: We should use isinstance here instead of comparring the type.

gtmanfred:
```suggestion
    if isinstance(schema_type, str):
        schema_types = [schema_type]
    elif isinstance(schema_type, list):
```

PR #317, opened 2024-08-08
Make `raise_400_for_marshmallow_errors` raise again

kaiku:
Buried in #300 was a consequential change to `raise_400_for_marshmallow_errors` that changed its behavior from raising errors to returning them.

This affects >= 3.2.0.

This PR reverts this change as it seems unnecessary.

### DIFF
--- a/flask_rebar/utils/request_utils.py
+++ b/flask_rebar/utils/request_utils.py
@@ -9,7 +9,15 @@
 """
 import collections
 import copy
-from typing import Any, Dict, Iterator, List, Optional, Type, Union, overload
+from typing import Any
+from typing import Dict
+from typing import Iterator
+from typing import List
+from typing import NoReturn
+from typing import Optional
+from typing import Type
+from typing import Union
+from typing import overload
 
 import marshmallow
 from marshmallow import Schema
@@ -148,7 +156,7 @@ def normalize_schema(
 
 def raise_400_for_marshmallow_errors(
     errs: Dict[str, Any], msg: Union[str, messages.ErrorMessage]
-) -> errors.BadRequest:
+) -> NoReturn:
     """
     Throws a 400 error properly formatted from the given marshmallow errors.
 
@@ -157,7 +165,7 @@ def raise_400_for_marshmallow_errors(
     :raises: errors.BadRequest
     """
     if not errs:
-        return errors.BadRequest(msg=msg)
+        raise errors.BadRequest(msg=msg)
 
     copied = copy.deepcopy(errs)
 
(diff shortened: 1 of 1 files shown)


flask_rebar/utils/request_utils.py
```diff
@@ -157,15 +165,15 @@ def raise_400_for_marshmallow_errors(
     :raises: errors.BadRequest
     """
     if not errs:
-        return errors.BadRequest(msg=msg)
+        raise errors.BadRequest(msg=msg)
```
airstandley, line 168: If there's no errors this function should just return. [That is what it was previously doing before this bug was introduced](https://github.com/plangrid/flask-rebar/blob/6e1d8d9d81a5d314ec162fdda67ee7a9f3accc6c/flask_rebar/utils/request_utils.py#L115-L133).
```suggestion
        return
```

airstandley (changes requested):
Thanks for catching this!

While I strongly dislike this pattern of having the function raise an error instead of returning a `BadRequest` for the caller to raise, the change in #300 broke a public interface which we should definitely avoid outside a major version.

Needs one small change to return the behaviour to what it was prior to #300.

Would also be great to see a unit test validating ` raise_400_for_marshmallow_errors`

flask_rebar/utils/request_utils.py
kaiku: I think keeping this change actually makes things better.

This function is designed around `marshmallow.ValidationError`. I think that if we handle a `ValidationError` and this function is called with the props of the validation error, it should raise a `BadRequest`. Intuitively, by naming, and practically, thinking about a web request, this makes sense.

Instead, we have a case where if `errs` is empty/falsy, we don't do anything. The only caller in flask-rebar is `_get_data_or_400` that passes `errs=e.messages_dict` where e is a `ValidationError`. The ([marshmallow src](https://github.com/marshmallow-code/marshmallow/blob/ac13e0d2353fb34431b14d6a541862a6b89f65c9/src/marshmallow/exceptions.py#L51C9-L51C22)) will raise a `TypeError` if the underlying `messages` prop isn't a dict, which tells me that it is, practically speaking, nearly or always set as a dict when raised as the result of `compat.load`, because we don't see this happening in production.

I think this is a case where introducing a bit of breakage in service of correctness is OK.

airstandley: I'm sold, as long as we document in the release notes that this is a bugfix where previously errors were not raised, it's a small enough change I'm on board with making it as a correction.

airstandley (approved):

Issue #183, opened 2020-04-23
OneOfConverter validator in array.


fcr--:
Given the following schema:
```python
class MyQuerySchema(QuerySchema):
    status = CommaSeparatedList(
        cls_or_instance=fields.Str(),
        validate=ContainsOnly(choices=["foo", "bar"]),
    )
```

Swagger-generator is producing the following output:
```javascript
                    {
                        "collectionFormat": "csv",
                        "enum": [
                            "foo",
                            "bar"
                        ],
                        "in": "query",
                        "items": {
                            "type": "string"
                        },
                        "name": "status",
                        "required": false,
                        "type": "array"
                    },
```

the problem here is that the `enum` property should be inside `items`, since enum (as defined in json schema) have the following defined behaviour: https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.2  that is to limit the possible values to those listed in its associated value.

If we were to have the enum property at the same level as we have the type array, we would need for this example to have this infinite array (and that's why we should instead place it inside items):
```javascript
{
  "type": "array",
  "enum": [
    [],
    ["foo"], ["bar"],
    ["foo", "foo"], ["foo", "bar"], ["bar","foo"], ["bar", "bar"],
    ["foo", "foo", "foo"], ...
  ]
  ...
}
```

auramix:
This was fixed, [here](https://github.com/plangrid/flask-rebar/pull/319#pullrequestreview-2355994325)

PR #319, opened 2024-10-02
[PJL-10221] handle some edge cases in OpenAPI v3.1 swagger generation

dekim24:
**Change 1**

Right now, ContainsOnly validators are being handled identically to OneOf validators. This is because, in Marshmallow, ContainsOnly is a subclass of OneOf. However, this isn't actually correct because the logic is different -

* ContainsOnly: *Validator which succeeds **if value is a sequence** and **each element in the sequence** is also in the sequence passed as choices.*
* OneOf: Validator which succeeds if **value** is a member of choices.

I added a separate validator to handle ContainsOnly and validated that it produces the expected output against one of our repos (PADD). You can see in the below screenshot that we were incorrectly specifying the `enum:` parameter at the level of the `value` itself (which causes an error if you run this against an OpenAPI validator, because you're comparing an array to a string) when the `enum:` should actually be at under the `items:`.

Screenshot 2024-10-02 at 8 58 13

**Change 2**

In OpenAPI 3 https://swagger.io/docs/specification/v3_0/data-models/dictionaries/ we are able to specify the allowed types of dictionary values (keys must always be strings so there is no need to specify their type) in the swagger file. The corresponding marshmallow param would be `values`. I added logic to factor in any schema passed into the `values` param and generate it in the swagger file. I also tested this on PADD and got expected results.

### DIFF
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -427,7 +427,7 @@ class DictConverter(FieldConverter[m.fields.Dict]):
     @sets_swagger_attr(sw.additional_properties)
     def get_additional_properties(
         self, obj: m.fields.Dict, context: _Context
-    ) -> Union[UNSET, m.fields.Dict]:
+    ) -> Union[Type[UNSET], m.fields.Dict]:
         if obj.value_field:
             return context.convert(obj.value_field, context)
         else:
--- a/flask_rebar/swagger_generation/marshmallow_to_swagger.py
+++ b/flask_rebar/swagger_generation/marshmallow_to_swagger.py
@@ -618,9 +618,7 @@ class ContainsOnlyConverter(ValidatorConverter):
     MARSHMALLOW_TYPE = ContainsOnly
 
     @sets_swagger_attr(sw.items)
-    def get_items(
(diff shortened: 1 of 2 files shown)


flask_rebar/swagger_generation/marshmallow_to_swagger.py
```diff
@@ -604,6 +614,16 @@ def get_maximum(
...
 
 
+class ContainsOnlyConverter(ValidatorConverter):
+    MARSHMALLOW_TYPE = ContainsOnly
+
+    @sets_swagger_attr(sw.items)
+    def get_items(
+        self, obj: ContainsOnly, context: _Context
+    ) -> Union[Type[UNSET], m.fields.List]:
+        return {"type": context.memo["items"]["type"], "enum": obj.choices}
```
dekim24, line 624: this is messy but I couldn't figure out a better way to figure out the `type` of the parent `value` when the `obj` we are passed is the `marshmallow.validator` i.e. 1 level inside the `value`.

auramix (approved):
Thanks for fixing this bug and contributing to flask-rebar, Eugene. Test cases look good-- happy to approve this.

I will add a link to this PR in the open issue from 2020 and mark it as closed: https://github.com/plangrid/flask-rebar/issues/183

Issue #320, opened 2024-12-05
HandlerRegistry.add_handler; authenticators


martinHrabina:
Hello everybody,

 due to [changes earlier this year](https://github.com/plangrid/flask-rebar/compare/v3.1.1...v3.2.0#diff-4d4a35db211c10faa00ffc4f32b1636c8448db808e0b23f776413dc0c7d5df2eR545), method `add_handler` of class `HandlerRegistry` silently ignores parameter `authenticators` if `Tuple[Authenticator]` is provided, resulting in authenticators being []. This is an issue especially when the value for this field comes from default value of another method, since lists (as mutable types) are discouraged and tuples are preferred in default values.

I'd expect either an error raised when unexpected type is received, and/or preferably to accept also tuples or iterables in general.

Ambrghani:
>

PR #322, opened 2024-12-11
Make authenticators accept Sequence

martinHrabina:
Addresses [issue  #320](https://github.com/plangrid/flask-rebar/issues/320), allows authenticator parameter to accept Sequence instead of just List and additionally does not ignore unsupported type but raises an error.

### DIFF
--- a/flask_rebar/rebar.py
+++ b/flask_rebar/rebar.py
@@ -492,7 +492,7 @@ class HandlerRegistry:
         request_body_schema: Optional[Schema] = None,
         headers_schema: Union[Type[USE_DEFAULT], Schema] = USE_DEFAULT,
         authenticators: Union[
-            Type[USE_DEFAULT], List[Authenticator], Authenticator
+            Type[USE_DEFAULT], Sequence[Authenticator], Authenticator
         ] = USE_DEFAULT,
         tags: Optional[Sequence[str]] = None,
         mimetype: Union[Type[USE_DEFAULT], str] = USE_DEFAULT,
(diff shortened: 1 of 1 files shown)


flask_rebar/rebar.py
```diff
@@ -547,12 +547,17 @@ def add_handler(
 
         # authenticators can be a list of Authenticators, a single Authenticator, USE_DEFAULT, or None
         authenticators_list: Sequence[Union[Type[USE_DEFAULT], Authenticator]] = []
-        if isinstance(authenticators, list):
-            authenticators_list = authenticators
+        if isinstance(authenticators, Sequence):
+            authenticators_list = list(authenticators)
```
nlawrence22, line 551: We're explicitly casting to a list here because we expect a list where this value is used?

flask_rebar/rebar.py
```diff
@@ -547,12 +547,17 @@ def add_handler(
...
         elif isinstance(authenticators, Authenticator) or authenticators is USE_DEFAULT:
             authenticators_list = [authenticators]
         elif authenticators is None:
             authenticators_list = []
+        else:
+            raise TypeError(
+                f"Unexpected data type '{type(authenticators)}' for parameter 'authenticators'. "
+                f"Expected type is either Authenticator, Sequence or None."
+            )
```
nlawrence22, line 560: 👍

nlawrence22 (commented):
Overall looks good, just one question I need to understand before approving. 👍

nlawrence22:
Also, thank you for the contribution!  Sorry, at first I thought this had come from my teammate -- we appreciate you taking the time to improve flask-rebar!

flask_rebar/rebar.py
martinHrabina: Honestly I am not really sure, in versions 3.1.1 and older, tuple was okay, but since 3.2.0 the check `isinstance(authenticators, list)` appeared for some reason. So I am casting to list here just to be sure (unsure if list is expected somewhere along the way or not)

PR #329, opened 2026-05-28
feat: openapi extension hooks for vendor extensions

gtmanfred:
## Summary

Adds two generic merge points to `SwaggerV3Generator` so callers can inject OpenAPI vendor extensions (`x-*` keys) without subclassing the generator or post-processing the spec dict, plus a small `openapi_extension` / `openapi_extensions` decorator pair as the supported API surface.

* **Per-operation hook** — the generator reads `getattr(path_def.func, "__openapi_extensions__", {})` and deep-merges it into the operation object emitted for that view.
* **Spec-root hook** — the generator reads `getattr(registry, "swagger_extensions", {})` and deep-merges it into the final spec dict before serialization.
* **Decorators** — `openapi_extension(key, value)` and `openapi_extensions(**kwargs)` stash on `func.__openapi_extensions__`, validate the `x-` prefix, and compose cleanly across multiple decorations.

A new `deep_merge` helper in `generator_utils` is the underlying primitive.

## Motivation

Downstream services need to emit Apigee-side vendor extensions in their generated OpenAPI specs (e.g. `x-rate-limit`, `x-internal`, plus vendor-specific keys downstream libraries can build typed wrappers around). Today those services hand-edit the generated spec or maintain a parallel spec. With these hooks, a downstream library can offer typed decorators / config that populate the attributes, and the generator merges them in for free.

The hooks are deliberately generic — rebar carries no knowledge of any specific vendor extension. Vendor-specific schemes (e.g. an OAuth2 clientCredentials scheme with an embedded `x-vendor-scopes` block) go through the existing `AuthenticatorConverter` path, demonstrated in the new test.

## Test plan

- [x] New tests under `tests/swagger_generation/test_openapi_extensions.py` cover both hooks, both decorators, the no-op-on-absent-attribute case, deep-merge semantics, override behavior on standard keys, decorator composition, the `x-` prefix validation, and the securityScheme-via-converter pattern.
- [x] Full `tests/swagger_generation/` suite passes (51 pass / 1 skip).
- [x] Manual: companion flask-toolbox change (https://github.com/plangrid/flask-toolbox/pull/311) builds typed decorators + a typed `ExtraSpecs` extension + an `OAuth2ClientCredentialsAuthenticator` on top, and the new spec generation has been trialed in two downstream services where the generated spec matches the hand-edited Apigee spec on every vendor-extension key.

## Draft

Draft until the companion flask-toolbox change lands. Merging this PR first is fine since the hooks are no-ops without callers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/swagger_generation/test_openapi_extensions.py
```diff
@@ -0,0 +1,251 @@
...
+
+    assert op["x-rate-limit"] == {"per-minute": 100}
+    assert "operationId" in op
+    assert "responses" in op
+
+
+def test_openapi_extensions_decorator_multiple_keys():
+    registry = _make_registry()
+
+    @registry.handles(rule="/foos", method="GET")
```
illusional, line 58: What's the difference between putting it on the handles vs a separate decorator if they're coupled together?

flask_rebar/swagger_generation/swagger_generator_v2.py
```diff
@@ -339,6 +344,10 @@ def _get_paths(
                 if d.tags:
                     path_definition[method_lower][sw.tags] = d.tags
 
+                func_extensions = getattr(d.func, "__openapi_extensions__", None)
```
illusional, line 347: Do you want to define `__openapi_extensions__` as a constant, or have some unified way of getting / setting it?

illusional (approved):

PR #333, opened 2026-08-26
Add support for pydantic models

illusional:
- Add `PydanticSchema` adapter to allow specifying Pydantic models as request (body, query, header) and response models in Flask Rebar
- Adds thin pydantic.BaseModel subclasses to replicate existing flask-rebar behaviour (extra=forbid, from_attribute=True, populate_by_name=True), and a camelCase helper subclass.
- Integrate this into flask-rebar (by checking if pydantic model, and auto-transforming to the adapter)
- Add pydantic as optional dependency (enable for tests)
- Adds documentation for this.

Note:
- responses are always validated, and this cannot be turned off


flask_rebar/utils/pydantic.py
```diff
@@ -0,0 +1,338 @@
...
+    def load(
+        self,
+        data: Any,
+        *,
+        many: bool | None = None,
+        partial: Any = None,
+        unknown: str | None = None,
+    ) -> BaseModel:
+        """Validate data with Pydantic and return its model instance."""
+        del many, partial
```
gtmanfred: What is this doing?

flask_rebar/utils/pydantic.py
```diff
@@ -0,0 +1,338 @@
...
+    # No return annotation: Pydantic uses it to build the *serialization* JSON Schema,
+    # and an explicit type here (even `Any`) breaks introspection for wrap-mode-serializers.
+    def _omit_none(self, handler: Any, info: SerializationInfo):  # type: ignore[no-untyped-def]
+        data = handler(self)
+        if not _omit_none_active(info):
+            return data
+        return {key: value for key, value in data.items() if value is not None}
+
+
+_OMIT_NONE = "flask_rebar_omit_none"
```
gtmanfred: constants should be at the top of the file.

flask_rebar/utils/pydantic_schema.py
```diff
@@ -0,0 +1,338 @@
...
+
+def validated_body(model: type[ModelType]) -> ModelType:
+    """Return Flask-Rebar's validated request body with its Pydantic type."""
+    del model
+    return cast(ModelType, get_validated_body())
+
+
+def validated_args(model: type[ModelType]) -> ModelType:
+    """Return Flask-Rebar's validated query arguments with their Pydantic type."""
+    del model
```
gtmanfred, line 269: What is this doing?

flask_rebar/utils/pydantic.py
```diff
@@ -0,0 +1,338 @@
...
+        for part in location[:-1]:
+            target = target.setdefault(part, {})
+        target.setdefault(location[-1], []).append(item["msg"])
+    return errors
+
+
+_SCHEMA_CACHE: dict[type[BaseModel], PydanticSchema] = {}
+_OPENAPI_SCHEMA_CACHE: dict[tuple[type[BaseModel], str], dict[str, Any]] = {}
+_INVALID_NAME_CHARS = re.compile(r"[^A-Za-z0-9._-]")
+_PROPERTY_MAP_KEYS = ("properties", "patternProperties", "$defs")
```
gtmanfred: can we move these to the top of the file?

gtmanfred (commented):
i just have a couple of requests and questions.

gtmanfred (commented):
## Review: pydantic support

Nice piece of integration work. I traced the diff against the current `master` call sites (`rebar.py`, `request_utils.py`, `compat.py`, `swagger_generation/*`) rather than just reading it in isolation. Findings below are grouped by severity.

Up front: the Major items are **code-trace conclusions, not executed repros** — I did not run the branch. #1 and #2 in particular should be easy to confirm or refute with a short test; happy to be wrong on either.

---

## Major

### 1. Cached `PydanticSchema` instances are mutated in place → cross-endpoint validation leakage

`schema_for()` in `flask_rebar/utils/pydantic.py` caches **one singleton `PydanticSchema` per model class**, regardless of the role that model is used in. But `get_header_params_or_400` in `flask_rebar/utils/request_utils.py` does:

```python
schema = compat.exclude_unknown_fields(schema)   # sets schema.unknown = marshmallow.EXCLUDE, in place
```

That permanently mutates the shared cached object on every request.

The plain-marshmallow path avoids this: `normalize_schema()` does `schema = schema()` for schema *classes*, so every handler registration gets its own instance. The pydantic path breaks that isolation invariant.

Failure scenario — reuse one model in two roles:

```python
class Paging(ApiModel):
    limit: int

@registry.handles(rule="/a", method="GET", headers_schema=Paging)
def a(): ...

@registry.handles(rule="/b", method="GET", query_string_schema=Paging)
def b(): ...
```

Both resolve to the same `schema_for(Paging)` object. The first request to `/a` sets `.unknown = EXCLUDE` on it permanently. From then on `/b`'s query-string validation silently **ignores** unknown fields instead of rejecting them per `extra="forbid"`. Request-order-dependent, silent.

Suggested fix: stop routing this through a mutable shared attribute. `PydanticSchema.load()` already accepts an `unknown` parameter — have `get_header_params_or_400` pass `unknown=marshmallow.EXCLUDE` per-call instead.

### 2. No `many` support — no list responses or array request bodies

```python
def load(self, data, *, many=None, partial=None, unknown=None):
    del many, partial          # accepted but ignored

def dump(self, obj, *, many=None):
    if many:                   # never populated by any caller
        return [self.dump(item) for item in obj]
```

Neither `marshal()`/`compat.dump()` nor `get_json_body_params_or_400()` pass `many=`, and `PydanticSchema` never consults `self.many` the way a marshmallow schema does. Since `schema_for()` returns one cached instance per model with no way to flip `.many`, there's currently no pydantic equivalent of `WidgetSchema(many=True)` — no array response, no JSON-array request body. `get_ref_schema()`'s `if not schema.many` branch also never fires for a pydantic-backed schema.

Failure scenario: a `GET /widgets` handler returning a list hits `dump()`'s single-object path, `model_validate(obj)` gets a list, and raises an opaque pydantic error at request time.

Suggested fix: either wire `many` through properly (likely needs a non-cached per-declaration wrapper for the many case), or raise `NotImplementedError` at **registration** time so it fails loudly and early rather than confusingly at request time.

### 3. Invalid Swagger 2 output with no runtime signal

`docs/pydantic.rst` L176-178 acknowledges pydantic models don't emit valid Swagger 2 — but `SwaggerV2Generator` is the **default** generator (`flask_rebar/rebar.py:387`) and nothing in code warns.

A team adopting `flask-rebar[pydantic]` without reading this specific new docs page gets an apparently-successful `generate_swagger()` producing a document with JSON-Schema-2020-12 constructs (`anyOf`, `"type": ["string","null"]`) that aren't valid Swagger 2.0 / draft-4 — and only finds out when a downstream validator or codegen tool chokes.

Suggested fix: `warnings.warn(...)` the first time a `PydanticSchema` is converted through a V2 generator, so the failure surfaces at swagger-generation time.

### 4. Component-name collisions silently corrupt the definitions dict

`component_name()` keys definitions purely on the sanitized class `__name__`, and `_flatten()` in `generator_utils.py` (L170-173) is last-write-wins with no collision check.

Existing marshmallow apps conventionally suffix schema classes (`WidgetResponseSchema`), which incidentally avoids collisions. Plain pydantic models are much more likely to reuse short generic names (`Item`, `Error`, `Nested`) across modules. Two distinct same-named models → one definition silently overwrites the other, and a `$ref` resolves to the wrong schema with no warning.

Suggested fix: detect the collision (raise or warn when `definitions[key]` already exists and differs). Arguably a pre-existing `_flatten()` issue, but pydantic usage makes it far likelier to bite, so worth handling here — e.g. namespace `component_name` with the module path on collision.

### 5. Recursive models silently degrade to a bare `{"type": "object"}`

`_inline_refs()`:

```python
if reference in stack:
    return {sw.type_: sw.object_}   # cycle: give up, no structure
```

Reasonable circuit breaker, but any self-referential model (`Node` with `children: list["Node"]` — common in pydantic) loses all nested structure at the recursion point, with no `$ref`, no warning, and no test. Not currently mentioned in the docs either.

Suggested fix: use flask-rebar's own `flatten()`/definitions mechanism, which already knows how to emit real `$ref`s, instead of fully inlining refs and then discarding cycles. At minimum, document the limitation.

---

## Minor

**6. `compat.dump()` dispatch consistency.** The new `getattr(schema, "dump_validates", False)` marker works and keeps pydantic out of `compat.py` — but the adjacent `RequireOnDumpMixin` check a few lines below uses `isinstance` for the same kind of dispatch. Consider a small marker class / `Protocol` for consistency and discoverability.

**7. `dump()` doesn't convert pydantic errors.** Unlike `load()`, `dump()` has no `try/except` around `model_validate`, so a response-marshaling failure raises a raw `pydantic.ValidationError`. It's still caught by the catch-all handler (`rebar.py:907`) → generic 500, so not a crash — but response failures don't get the `rebar_error_code`/structured-error treatment request failures do, and under `app.debug=True` surface a raw pydantic exception. Worth a line in the docs' Responses note, which currently says validation "will raise" without saying in what shape.

**8. `extra="forbid"` on query strings may be too strict.** The docs carve headers out of `extra="forbid"` (correctly — every request carries incidental headers) but query strings get no such treatment. Any unexpected param (`utm_source`, cache-busters, gateway-injected params) will 400. Consistent with the code as written, but a sharp edge worth documenting, since reaching for `ApiModel` on a query string is the obvious default.

**9. Silent key-dropping on response dump can mask handler bugs.** `dump()` filters mappings down to `self._known_keys` — intentional, and explicitly asserted in `test_a_plain_dict_response_is_validated_and_coerced_through_the_model`. But it means a typo in an *optional* field name in a returned dict produces no error at all; the data just vanishes from the response. Worth a docs callout next to the "always validates" note, since it's the opposite failure mode from request-side `extra="forbid"`.

**10. Test coverage gaps.** The 283-line test file is well organized, but given the scope I'd flag: no end-to-end test against `SwaggerV3Generator` (the docs recommend V3, but only the default V2 path is exercised); nothing covering the cross-role `.unknown` mutation in #1; nothing for `many` (#2), recursion (#5), or name collisions (#4); and no test exercising a response-marshaling *failure* to pin down the resulting status and body shape.

---

## Nit

**11. Module name `flask_rebar/utils/pydantic.py`.** I checked this specifically — Python 3's absolute imports mean `from pydantic import BaseModel` inside that file correctly resolves to the real package, so there's **no actual shadowing bug**. Still a readability/tooling nit (grep noise, autocomplete confusion), and the PR already introduces a differently-named `pydantic_helpers.py`. Consider `pydantic_schema.py`.

**12. Unbounded module-level caches.** `_SCHEMA_CACHE` and `_OPENAPI_SCHEMA_CACHE` are plain dicts with no eviction. Fine when models are defined at import time, but would grow unbounded if an app constructs model classes dynamically (parametrized generics). Worth a doc note that models should be defined at import time.

---

## What this PR does well

- **Clean optional-dependency boundary.** `pydantic_helpers.py` soft-imports pydantic and is safe to import unconditionally from `request_utils.py`, while the real adapter hard-imports but is only reached lazily and raises a clear `ImportError` with an install hint. Verified by tracing `flask_rebar/__init__.py → request_utils.py → pydantic_helpers.py` — no pydantic dependency leaks onto the base install path.
- **Clever reuse of the existing swagger machinery.** Rather than building a parallel pipeline, the adapter registers `PydanticSchema` with the request-body/response converter registries for full nested schemas, while relying on the *existing* generic `SchemaConverter` + `flatten()` for query/header params by pre-populating `schema.fields` with lightweight `_JsonSchemaField` placeholders. That gets correct `required`/`additionalProperties` behavior for free from code that already exists. Well thought out.
- **Explicit, scoped version support.** Pinning `pydantic>=2.10,<3` and committing to v2-only APIs (`ConfigDict`, `model_validate`, `model_json_schema`) instead of straddling v1/v2 is the right call for a first cut, and it's clearly documented.
- **No dynamic-code-execution risk.** No `eval`, `exec`, or unsafely-driven `create_model` anywhere in the diff.
- **Thoughtful query-string handling.** `_flatten_multidict` using `model_fields` annotation introspection (`_accepts_sequence`) to decide list-vs-scalar is a nice low-boilerplate way to support repeated `?tags=a&tags=b`, and it's actually tested.
- **Good-faith documentation of known limitations.** The Swagger 2 incompatibility and the always-validate response behavior are called out rather than hidden. I'd mainly like a couple of them backed by a runtime signal (#3) rather than prose alone.

flask_rebar/utils/pydantic_schema.py
illusional: Explicitly not using the variable without getting linting problems

illusional:
Thanks @gtmanfred, 

1. Cached PydanticSchema instances are mutated in place → cross-endpoint validation leakage: I hadn't realized there was internal mutation, was trying to save a few compute cycles. I've removed the top level cache, and it'll just instantiate an instance for each type you specify it.
2. No many support — no list responses or array request bodies: I was meh on this originally, but it can block the case where your body is just straight `list[T]`. I've made it support `many=True`, and the pydantic.RootModel trick. There are tests and docs for both
3. Invalid Swagger 2 output with no runtime signal: Don't know why you couldn't access the context to check the version, added a warning. Happy to make that an error too.
4. Component-name collisions silently corrupt the definitions dict: Added a specific error for this.
5. Recursive models silently degrade to a bare {"type": "object"}: Great catch, added some test cases and fixed this.
6. compat.dump() dispatch consistency: Ignored
7. `.dump()` doesn't convert pydantic errors: Instead of converting to marshmallow validation errors, I've changed this more dramatically to raise pydantic errors directly - and made request_utils handle these. IE: Use pydantic models, get pydantic shaped results. I realize this is a departure, but part of the ergnonomics of pydantic is the error system. 
8. extra="forbid" on query strings may be too strict: Okay yep I changed the guidance here in the pydantic docs
9. Silent key-dropping on response dump can mask handler bugs: IMO this is okay, I'd recommend people return the pydantic object they're actually wanting to return rather than a dictionary. You have the pydantic model, use it.

gtmanfred (approved):

